PR #140, opened 2026-04-09
feat: add fheuint/bdd tmp bytes in poulpy-schemes

### LINKED ISSUE #139: Scratch sizing API for `FheUint` BDD ops and encrypt/decrypt
`poulpy-schemes` uses `Scratch` arenas for hot paths. Callers are expected to allocate enough space up front. Several `FheUint` flows (word-level BDD evaluation after prepare, and encrypt/decrypt) did not expose a supported way to compute that size, so integrators had to guess, over-allocate, or duplicate internal layout logic.

Poulpy could provide `*_tmp_bytes` (or equivalent) entry points that return a safe lower bound for scratch usage for those missing operations, consistent with other APIs in the crate (e.g. prepare / key material).

Then from public APIs only, callers should be able to choose a scratch size that is sufficient for the intended operations (e.g. [here](https://github.com/cedoor/squid/blob/11964c5f3015a497fd2c66112f5f52c6627bd184/src/context.rs#L521))

Pls correct me if I got something wrong.

cedoor:
Closes #139

### DIFF
--- a/poulpy-schemes/src/bin_fhe/bdd_arithmetic/tests/test_suite/add.rs
+++ b/poulpy-schemes/src/bin_fhe/bdd_arithmetic/tests/test_suite/add.rs
@@ -11,8 +11,7 @@ use rand::Rng;
 
 use crate::bin_fhe::{
     bdd_arithmetic::{
-        Add, BDDKeyEncryptSk, BDDKeyPrepared, BDDKeyPreparedFactory, ExecuteBDDCircuit2WTo1W, FheUint,
-        FheUintPrepare,
+        Add, BDDKeyEncryptSk, BDDKeyPrepared, BDDKeyPreparedFactory, ExecuteBDDCircuit2WTo1W, FheUint, FheUintPrepare,
         FheUintPrepareDebug, FheUintPrepared, FheUintPreparedEncryptSk, FheUintPreparedFactory,
         tests::test_suite::{TEST_GGSW_INFOS, TEST_GLWE_INFOS, TestContext},
     },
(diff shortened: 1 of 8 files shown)


poulpy-schemes/src/bin_fhe/bdd_arithmetic/tests/test_suite/add.rs
```diff
@@ -63,8 +63,22 @@ where
     source.fill_bytes(&mut scratch.borrow().data);
     b_enc_prep.encrypt_sk(module, b, sk_glwe_prep, &mut source_xa, &mut source_xe, scratch.borrow());
 
-    // a + b
-    res.add(module, &a_enc_prep, &b_enc_prep, bdd_key_prepared, scratch.borrow());
+    let atk_infos = bdd_key_prepared.automorphism_key_infos();
+    let add_bytes: usize = res.add_tmp_bytes(module, &glwe_infos, &ggsw_infos, &atk_infos);
```
Pro7ech: I find it weird that atk needs to be passed there. Atk keys are accessible, but one shoudn't have to interact with it in regular usage of the scheme. Probably its better to ask for the bdd key.

poulpy-schemes/src/bin_fhe/bdd_arithmetic/tests/test_suite/fheuint.rs
```diff
@@ -179,3 +179,29 @@ where
         assert_eq!(a.bit(i) as u32, c_enc.decrypt(module, sk, scratch.borrow()));
     }
 }
+
+pub fn test_fhe_uint_encrypt_decrypt_tmp_bytes<BRA: BlindRotationAlgo, BE: Backend>(test_context: &TestContext<BRA, BE>)
```
Pro7ech: I'm not for having a specific test dedicated the scratch space. I'd rather have it done directly in the encrypt decrypt original test method using a separate scratch space allocated with just what enc/dec for the enc/dec calls.

poulpy-schemes/src/bin_fhe/bdd_arithmetic/tests/test_suite/add.rs
cedoor: Oh makes sense!

Pro7ech (approved):

PR #144, opened 2026-04-14
hal refactor

Pro7ech:
cf changelog

PR #147, opened 2026-04-16
feat(serialization): ReaderFrom/WriterTo for LWE and BDD keys

### LINKED ISSUE #145: `ScalarZnx::write_to` includes alignment padding, breaking roundtrips
https://github.com/poulpy-fhe/poulpy/blob/5d0df23b2e5a92f81ce18be93a76084af726a523/poulpy-hal/src/layouts/scalar_znx.rs#L318-L327

`write_to` is `writing buf.len()` (which may include alignment padding) while `read_from` expects exactly `n * cols * 8` bytes.

This should cause roundtrip failures when the backing buffer is over-allocated (?).

### LINKED ISSUE #146: Add `ReaderFrom`/`WriterTo` for `LWESecret`, `CircuitBootstrappingKey`, and `BDDKey`
These key types lack serialization support via the existing `ReaderFrom`/`WriterTo` pipeline. This blocks persisting/loading full evaluation key bundles without going through a higher-level wrapper. The implementation should cover:

- `LWESecret<D>` (distribution tag + `ScalarZnx` data)
- `CircuitBootstrappingKey<D, BRA>` (brk + sorted atk HashMap + tsk)
- `BDDKey<Vec<u8>, BRA>` (cbt + optional ks_glwe with tag + ks_lwe)

cedoor:
Fixes #145 
Closes #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,15 @@
 ## [Unreleased] - 2026-04-17
 
 ### `poulpy-hal`
-- Make `WriterTo` for `MatZnx`, `ScalarZnx`, `VecZnx`, and `SvpPPol` emit the canonical logical byte length from layout metadata (or `B::bytes_of_svp_ppol`), write only that prefix, and error when backing storage is shorter than the logical span.
-- Tighten `SvpPPol` `ReaderFrom` by requiring the on-wire payload length to match `bytes_of_svp_ppol` for the decoded `n` and `cols`.
+- Make `WriterTo` for `MatZnx` and `VecZnx` emit the canonical logical byte length from layout metadata, write only that prefix, and error when backing storage is shorter than the coefficient span.
+- Fix `ScalarZnx::write_to` to emit the full `n * cols` coefficient byte span (aligned `i64` layout).
+- **Breaking:** Remove `ReaderFrom` / `WriterTo` for prepared DFT layouts (`SvpPPol`); remove `SvpPPolFromBytes`, `VmpPMatFromBytes`, and `from_bytes` on the corresponding prepared types. Document that `SvpPPol` / `VmpPMat` DFT alignment assumes a power-of-two ring degree.
 
 ### `poulpy-schemes`
-- Add `ReaderFrom` / `WriterTo` for `CircuitBootstrappingKey` and for `BDDKey<Vec<u8>>` (tagged optional `ks_glwe`), with deterministic ATK map serialization via sorted Galois keys.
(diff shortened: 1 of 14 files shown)


Pro7ech:
- All good for the public material serialiation also #145 is also an issue for VecZnx (which is used for LWECiphertext), so VecZnx serialization also needs to be updated.
- After giving it some though, I think providing serialization for secrets, was a mistake. If someone wants to transfer such secret, it should do it outside of the library, by sending the seed that was used to generate the lattice secret. So I suggest removing all serialization of secret from the library.
- Also please update the changelog accordingly.

cedoor:
> All good for the public material serialiation also [`ScalarZnx::write_to` includes alignment padding, breaking roundtrips #145](https://github.com/poulpy-fhe/poulpy/issues/145) is also an issue for VecZnx (which is used for LWECiphertext), so VecZnx serialization also needs to be updated.

I think MatZnx and SvpPPol share the same issue. Can you confirm?

> After giving it some though, I think providing serialization for secrets, was a mistake. If someone wants to transfer such secret, it should do it outside of the library, by sending the seed that was used to generate the lattice secret. So I suggest removing all serialization of secret from the library.

That makes sense.

> Also please update the changelog accordingly.

Done with a unreleased block, same as [PR#144](https://github.com/poulpy-fhe/poulpy/pull/144/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed).

Pro7ech:
> I think MatZnx and SvpPPol share the same issue. Can you confirm?

- SvpPPol and VmpPMat are prepared structs, in the DFT domain, so they always have a power of two number of coefficients, which makes their lane size always a multiple of the alignment. **Also, there shouldn't be any way to serialize prepared structs, if there is it should be removed.**
- MatZnx, can be instantiated with a non-power of two size, which would cause the roundtrip problem. In practice MatZnx is a collection of GLWE, so the lane size is always a power of two, and thus a multiple of the alignment. But I still think that, to be consistent with VecZnx and ScalarZnx (which are user facing structs), applying the patch there makes sens.

cedoor:
> SvpPPol and VmpPMat are prepared structs, in the DFT domain, so they always have a power of two number of coefficients, which makes their lane size always a multiple of the alignment. **Also, there shouldn't be any way to serialize prepared structs, if there is it should be removed.**

Got it, I removed Reader/Writer and *from_bytes functions from both of them. Let me know if that makes sense. Otherwise I'll just revert my previous change.

Pro7ech (approved):

PR #154, opened 2026-04-26
Add ckks

Pro7ech:
cf changelog


poulpy-ckks/README.md
```diff
@@ -0,0 +1,243 @@
...
+  granular at the bit level, evaluation keys are not tied to a specific level
+  schedule or prime decomposition for a given circuit.
+
+The goal of this representation is not just ergonomics. It is meant to provide
+those advantages while remaining comparable in performance to state-of-the-art
+RNS CKKS libraries.
+
+Each ciphertext carries CKKS metadata:
+
+- `log_decimal`: base-2 logarithm of the plaintext precision
```
jdumezy: Terminology needs to be updated

poulpy-ckks/README.md
```diff
@@ -0,0 +1,243 @@
...
+- `CKKSPlaintextVecRnx<F>`
+- `CKKSPlaintextCstRnx<F>`
+- `CKKSPlaintextCstZnx`
+- `CKKSMeta`
+
+`CKKSMeta` stores the logical precision metadata used by the scheme:
+
+```rust
+pub struct CKKSMeta {
+    pub log_decimal: usize,
```
jdumezy: Same terminology issue

jdumezy (dismissed):

Janmajayamall:
Overall the PR LGTM. I asked my agent for a consistency review and it came up with a few points that I'd like to consider.

---

## Bug 1 — `ckks_div_pow2_into_default` does not apply the `bits` shift to limb data

**File:** `poulpy-ckks/src/leveled/default/pow2.rs:69`

Line 69 calls:
```rust
self.glwe_lsh(dst, src, offset, scratch);
```
but line 71 reduces `log_budget` by `bits + offset`:
```rust
dst.meta.log_budget = checked_log_budget_sub("div_pow2", dst.log_budget(), bits + offset)?;
```

The metadata says the ciphertext was divided by `2^bits`, but the actual limb data was only aligned by `offset` — the `bits` component is never applied to the data. Compare `ckks_mul_pow2_into_default` (line 37), which correctly passes `bits + offset` to `glwe_lsh`.

If `div_pow2` is intentionally a pure metadata operation (semantic-only scale change with no data shift), that decision needs explicit documentation and the symmetric `mul_pow2` behaviour needs to be reconciled. As written, the two operations are not inverses of each other.

---

## Bug 2 — `ckks_add_assign_unsafe_default` does not update `log_delta` when operands differ

**File:** `poulpy-ckks/src/leveled/default/add.rs:143`

Line 143 only updates `log_budget`:
```rust
dst.meta.log_budget = dst_log_budget.min(a.log_budget());
```
`log_delta` is never touched. The safe `_into` counterpart (line 102) correctly takes:
```rust
dst.meta.log_delta = a.log_delta().max(b.log_delta());
```
If `dst.log_delta() < a.log_delta()`, the resulting ciphertext carries a stale precision tag. Any downstream operation that reads `log_delta` for noise estimation or rescaling alignment will silently produce incorrect results. The same omission exists in `ckks_sub_assign_unsafe_default`.

Fix:
```rust
dst.meta.log_delta = dst.log_delta().max(a.log_delta());
dst.meta.log_budget = dst_log_budget.min(a.log_budget());
```

---

## Bug 3 — `mul_many_rec` assumes uniform `log_delta` across all inputs without validation

**File:** `poulpy-ckks/src/leveled/delegates/composite.rs:129`

Line 129 reads `log_delta` from `inputs[0]` only and applies it to all scratch-size estimates (lines 132–133):
```rust
let log_delta: usize = inputs[0].log_delta();
let left_max_k = left_min_eff_k.saturating_sub(ceil_log2(left_slice.len()) * log_delta);
let right_max_k = right_min_eff_k.saturating_sub(ceil_log2(right_slice.len()) * log_delta);
```
With mixed `log_delta` inputs the scratch buffers are mis-sized and the output metadata will be wrong. Neither `ckks_mul_many` nor this function documents or enforces the uniform-`log_delta` invariant.

At minimum, add a validation guard:
```rust
ensure!(
    inputs.iter().all(|c| c.log_delta() == inputs[0].log_delta()),
    "ckks_mul_many: all inputs must have the same log_delta"
);
```
or use `inputs.iter().map(|c| c.log_delta()).max().unwrap()` as a conservative upper bound.

---

## Security — `pub meta` allows external callers to silently corrupt scheme invariants

**File:** `poulpy-ckks/src/layouts/ciphertext.rs:22`  
(same applies to `poulpy-ckks/src/layouts/plaintext/vec.rs:28`)

Both `CKKSCiphertext.meta` and `CKKSPlaintextVecZnx.meta` are fully `pub`, so any downstream crate can write:
```rust
ct.meta.log_budget = 999;
ct.meta.log_delta = 0;
```
without going through any checked path. For a cryptographic library where these fields directly control noise-budget accounting and output precision, unchecked external mutation is a meaningful footgun.

If the goal is to allow test-internal mutation (as in `errors.rs`), consider making the fields `pub(crate)` and exposing deliberate escape hatches (`#[cfg(test)] pub` or a checked setter) for the few places that need them. If they are intentionally `pub` for advanced library users, add a prominent doc comment explaining the invariants callers are responsible for maintaining.

---

## Security — Panic instead of propagated `Result` on arithmetic overflow

**File:** `poulpy-ckks/src/leveled/default/add.rs:419` (also line 461, and symmetrically in `default/sub.rs`)

```rust
res_log_budget
    .checked_add(prec.log_delta)
    .expect("aligned precision overflow")
```
Every other failure path in the evaluator returns `anyhow::Result`. This path panics — unrecoverable and inconsistent with the rest of the API. Replace with:
```rust
res_log_budget
    .checked_add(prec.log_delta)
    .ok_or_else(|| anyhow::anyhow!("add_const_rnx: aligned precision overflow"))?
```

---

## Consistency — `CKKSMulAddOps` / `CKKSMulSubOps` methods are missing the `_into` suffix

**File:** `poulpy-ckks/src/leveled/api/composite.rs:88`

Every other operation family follows:
- `ckks_op_into(dst, a, b, ...)` — result written to a separate `dst`
- `ckks_op_assign(dst, a, ...)` — result accumulated into `dst`

`ckks_mul_add_ct`, `ckks_mul_add_pt_vec_znx`, `ckks_mul_sub_ct`, etc. take a `dst` + two operands (the `_into` pattern) but drop the suffix. This makes the API inconsistent and harder to search.

Suggested rename: `ckks_mul_add_ct_into`, `ckks_mul_sub_ct_into`, etc.

---

## Consistency — `alloc_pt_znx` is an undeprecated public duplicate of `alloc_pt_vec_znx`

**File:** `poulpy-ckks/src/layouts/plaintext/vec.rs:78`

```rust
pub fn alloc_pt_znx(n: Degree, base2k: Base2K, prec: CKKSMeta) -> CKKSPlaintextVecZnx<Vec<u8>> {
    alloc_pt_vec_znx(n, base2k, prec)  // identical body
}
```
Both are public, neither is marked `#[deprecated]`. Users have no signal about which to prefer and both appear in documentation and autocomplete.

Either remove `alloc_pt_znx`, or mark it deprecated:
```rust
#[deprecated(since = "0.5.0", note = "use alloc_pt_vec_znx instead")]
pub fn alloc_pt_znx(...)
```

---

## Tests — `FFT64_PARAMS_F64.log_budget = 0` skips all capacity-consuming operations

**File:** `poulpy-ckks/src/leveled/tests/test_suite/mod.rs:86`

```rust
prec: CKKSMeta { log_delta: 30, log_budget: 0 },
```
Any operation that consumes even 1 bit of budget — multiplication, squaring, const-add, rescaling — will fail at the `checked_log_budget_sub` guard before executing any FFT64 arithmetic. The full test suite macro is still instantiated against this parameter set, so those tests exist but provide no backend coverage.

If `log_budget: 0` is intentional for this backend's constraints, the test suite instantiation should be narrowed to only the operations that are valid (e.g. encrypt/decrypt, addition). Otherwise, raise `log_budget` to a value that exercises the full suite (e.g. `30`, matching `NTT120_PARAMS_F64`).

---

## Minor — Guard order in `Encoder::new` causes `m = 0` to emit a misleading error

**File:** `poulpy-ckks/src/encoding/reim.rs:35`

```rust
anyhow::ensure!(m.is_power_of_two(), "m must be a power of two, got {m}");
anyhow::ensure!(m > 0, "m must be > 0, got {m}");
```
In Rust, `0_usize.is_power_of_two()` returns `false`, so `m = 0` is caught by the first guard with the message _"m must be a power of two"_ — the correct error is never reached.

Swap the guards:
```rust
anyhow::ensure!(m > 0, "m must be > 0, got {m}");
anyhow::ensure!(m.is_power_of_two(), "m must be a power of two, got {m}");
```

---

## Minor — `checked_mul_pt_log_budget` is not re-exported consistently with its siblings

**File:** `poulpy-ckks/src/error.rs:177`

All other `pub(crate)` helpers in this file are re-exported at the crate root in `lib.rs`:
```rust
pub(crate) use error::{
    checked_log_budget_sub,
    ensure_base2k_match,
    ensure_plaintext_alignment,
    checked_mul_ct_log_budget,
    // checked_mul_pt_log_budget  <-- missing
};
```
`checked_mul_pt_log_budget` must be referenced as `crate::error::checked_mul_pt_log_budget` while all its siblings are available as `crate::checked_*`. Add it to the re-export block in `lib.rs` for consistency.

Janmajayamall (approved):

PR #165, opened 2026-05-21
Dev poly eval 

Pro7ech:


jdumezy (dismissed):

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    pub fn chebyshev_interpolate<Fun>(degree: usize, a: F, b: F, f: Fun) -> Result<Self>
+    where
+        Fun: Fn(F) -> F,
+    {
+        chebyshev_interpolate(degree, a, b, f)
+    }
+
+    pub fn degree(&self) -> usize {
```
Janmajayamall: Medium: `Polynomial::degree()` does not trim trailing zero coefficients

This treats trailing zero coefficients as part of the degree. For example, `vec![1.0, 2.0, 0.0, 0.0]` is treated as degree 3 instead of degree 1, which can cause unnecessary BSGS decomposition, power-basis generation, encoded plaintext size, and ciphertext multiplications.

Suggestion: trim trailing zeros in `Polynomial::new`, or add/use an `effective_degree()` helper in `encode_bsgs_with`.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+            .all(|(i, &c)| !i.is_multiple_of(2) || c == F::zero())
+        {
+            Parity::Odd
+        } else {
+            Parity::Full
+        };
+        Self::new_with_parity(basis, coeffs, parity)
+    }
+
+    pub fn new_with_parity(basis: Basis, coeffs: Vec<F>, parity: Parity) -> Self {
```
Janmajayamall: Medium: `new_with_parity` can silently create incorrect polynomials

This constructor trusts the caller. If a caller marks a full polynomial as `Parity::Even` or `Parity::Odd`, evaluation will skip coefficients and return an incorrect result.

Suggestion: validate parity in debug builds, provide a checked constructor, or clearly document that incorrect parity causes incorrect evaluation.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+        &self,
+        module: &Module<HostBytesBackend>,
+        base2k: Base2K,
+        coeff_meta: CKKSMeta,
+        strategy: SplitStrategy,
+    ) -> Result<BSGSPolynomial<CKKSPlaintext<Vec<u8>>>>
+    where
+        F: crate::layouts::CKKSScalar,
+    {
+        ensure!(self.degree() >= 1, "polynomial must have degree ≥ 1");
```
Janmajayamall: Low/Medium: constant polynomials are rejected by `encode_bsgs_with`

This rejects degree-0 constant polynomials. That may be intentional, but constants are common and could be represented as a single baby step without requiring power-basis generation.

Suggestion: either support degree-0 polynomials or explicitly document this limitation in the public API.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+            }
+        }
+    }
+
+    /// Evaluates this polynomial on an input interval.
+    ///
+    /// Monomial polynomials are evaluated directly at `x`. Chebyshev
+    /// polynomials first map `x` from `[a, b]` to the normalized Chebyshev
+    /// variable `(2x-a-b)/(b-a)`.
+    pub fn evaluate_on_interval(&self, x: F, a: F, b: F) -> F {
```
Janmajayamall: Low: `evaluate_on_interval` does not validate interval

Unlike `chebyshev_interpolate`, this does not check `a < b`; if `a == b`, it divides by zero.

Suggestion: return `Result<F>`, add validation, or at least add `debug_assert!(a < b)` for the Chebyshev path.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+    needed.len()
+}
+
+// ── Polynomial ───────────────────────────────────────────────────────────────
+
+/// A plaintext polynomial with real coefficients.
+///
+/// `coeffs[i]` is the coefficient of the degree-`i` term (monomial basis) or
+/// of `Tᵢ(x)` (Chebyshev basis).
+pub struct Polynomial<F> {
```
Janmajayamall: Medium: `Polynomial` fields are public, allowing parity invariants to be broken

Even if `Polynomial::new` detects parity correctly, callers can later mutate `coeffs` or `parity` independently and silently make evaluation incorrect.

Suggestion: make fields private and expose accessors, or clearly document that this is an invariant-sensitive struct.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+
+// ── BSGSPolynomial ────────────────────────────────────────────────────────────
+
+/// A polynomial decomposed for Baby-Step-Giant-Step (BSGS) evaluation.
+///
+/// `baby_steps[0]` is the lowest-degree encoded baby polynomial containing the constant and
+/// low-degree terms; `baby_steps[n−1]` is the highest-degree encoded baby polynomial.
+///
+/// Construct via [`Polynomial::encode_bsgs`].
+pub struct BSGSPolynomial<C> {
```
Janmajayamall: Low/Medium: `BSGSPolynomial` fields are public despite strong structural invariants

The evaluator assumes baby steps are ordered and shaped exactly as produced by `Polynomial::encode_bsgs`. External mutation can break these assumptions and cause incorrect evaluation or panics.

Suggestion: make fields private and provide accessors such as `degree()`, `base()`, `log_split()`, and `baby_steps()`.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+///
+/// `coeffs[i]` is the coefficient of the degree-`i` term (monomial basis) or
+/// of `Tᵢ(x)` (Chebyshev basis).
+pub struct Polynomial<F> {
+    pub basis: Basis,
+    pub coeffs: Vec<F>,
+    pub parity: Parity,
+}
+
+impl<F> Polynomial<F>
```
Janmajayamall: Low: generic bounds are broader than necessary for most `Polynomial` methods

Methods like `new`, `degree`, `evaluate`, and `encode_bsgs` do not all need `FloatConst`, `FromPrimitive`, or `Debug`. This unnecessarily restricts API compatibility.

Suggestion: split into smaller `impl` blocks, e.g. `F: Float` for construction/evaluation/parity; `F: Float + FloatConst + FromPrimitive` for Chebyshev interpolation; and CKKS-specific bounds only for encoding.

poulpy-ckks/src/polynomial.rs
```diff
@@ -0,0 +1,531 @@
...
+fn pt_ct_muls_for_baby_step(degree: usize, parity: Parity) -> usize {
+    let (first, step) = match parity {
+        Parity::Even => (2usize, 2usize),
+        Parity::Odd => (1, 2),
+        Parity::Full => (1, 1),
+    };
+    if degree < first { 0 } else { (degree - first) / step + 1 }
+}
+
+fn estimate_power_basis_muls(degree: usize, log_split: usize, parity: Parity, basis: Basis) -> usize {
```
Janmajayamall: Low/Medium: `estimate_power_basis_muls` can overestimate baby-step powers when `degree < base`

This estimates baby-step powers up to `base - 1`, even when the actual polynomial degree is smaller than `base - 1`. For example, `degree = 5`, `log_split = 3`, `base = 8`, `Parity::Full` counts powers through `X^7`, but a degree-5 baby polynomial only needs up to `X^5`.

Suggestion: cap baby targets with something like `let baby_limit = base.min(degree + 1);` and use that instead of `base` in the parity ranges. Apply the same idea in `PowerBasis::populate` if the intent is to avoid generating unused powers.

poulpy-ckks/src/power_basis.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.values.get(&n)
+    }
+
+    /// Inserts a pre-computed power at degree `n`.
+    pub fn insert(&mut self, n: usize, value: A) {
+        self.values.insert(n, value);
+    }
+}
+
+impl<BE: Backend, A> PowerBasisHelper<BE, A> for PowerBasis<A>
```
Janmajayamall: Medium/High: polynomial evaluation cannot validate `PowerBasis` basis compatibility

`PowerBasis` stores a `basis`, but `PowerBasisHelper` only exposes `get()`. Evaluation therefore cannot verify that a Chebyshev `BSGSPolynomial` is evaluated with a Chebyshev `PowerBasis`, or that a monomial polynomial is evaluated with a monomial `PowerBasis`. Since `get(i)` has basis-dependent meaning (`X^i` vs `T_i(X)`), a mismatch can silently return incorrect values.

Suggestion: add `fn basis(&self) -> Basis` to `PowerBasisHelper` and check `poly.basis() == power_basis.basis()` near the top of the evaluator.

poulpy-ckks/src/power_basis.rs
```diff
@@ -0,0 +1,269 @@
...
+    ) -> Result<()>
+    where
+        BE: Backend<OwnedBuf = D>,
+        Module<BE>: CKKSAddOps<BE> + CKKSMulOps<BE> + CKKSSubOps<BE> + CKKSModuleAlloc<BE>,
+        CKKSCiphertext<D>: GLWEToBackendMut<BE> + GLWEToBackendRef<BE> + CKKSCtBounds + SetCKKSInfos,
+        poulpy_core::layouts::GLWETensorKeyPrepared<D, BE>: GLWETensorKeyPreparedToBackendRef<BE> + GGLWEInfos,
+    {
+        ensure!(degree >= 1, "populate: degree must be ≥ 1");
+
+        let log_degree = (usize::BITS - degree.leading_zeros()) as usize;
```
Janmajayamall, line 222: Low/Medium: `populate` over-generates powers when `degree < base`

For `degree = 5`, `log_split = 3`, `base = 8`, and `Parity::Full`, this generates powers up to `7`, even though a degree-5 baby polynomial only needs powers up to `5`.

Suggestion: cap baby-step generation at the actual degree with something like `let baby_limit = base.min(degree + 1);` and use `baby_limit` instead of `base` in the parity ranges.

poulpy-ckks/src/power_basis.rs
```diff
@@ -0,0 +1,269 @@
...
+// ── PowerBasis ────────────────────────────────────────────────────────────────
+
+/// Stores pre-computed powers of a ciphertext for BSGS polynomial evaluation.
+///
+/// `values[n]` = X^n (monomial basis) or Tₙ(X) (Chebyshev basis).
+/// `values[1]` must be provided at construction time.
+///
+/// Implements [`PowerBasisHelper`] so it can be passed directly to
+/// `ckks_eval_poly_real_const_coeffs_from_power_basis`.
+pub struct PowerBasis<A> {
```
Janmajayamall: Medium: public `basis` plus unrestricted `insert` can break `PowerBasis` invariants

Callers can mutate `basis` after powers are inserted, or insert arbitrary ciphertexts that do not correspond to the declared basis/degree/metadata. This can silently produce incorrect polynomial evaluation.

Suggestion: make `basis` private and expose `basis()` as a read-only accessor. Consider restricting `insert`, documenting its invariants, or adding checked insertion helpers.

poulpy-ckks/src/power_basis.rs
```diff
@@ -0,0 +1,269 @@
...
+    let log_budget = checked_mul_ct_log_budget("power_basis", a.log_budget(), b.log_budget(), a.log_delta(), b.log_delta())?;
+    Ok(log_budget + a.log_delta().min(b.log_delta()))
+}
+
+/// Splits `n` into `(a, b)` with `n = a + b` and `|a – b|` minimised.
+///
+/// When `n` is a power of two `a = b = n/2`; otherwise uses the
+/// Lee et al. (2020) strategy that maximises the number of odd-degree
+/// Chebyshev terms.
+pub(crate) fn split_degree(n: usize) -> (usize, usize) {
```
Janmajayamall: Low: `split_degree(0)` and `split_degree(1)` are not guarded

Current callers use `split_degree` only after checking `n >= 2`, so this is not currently exposed through `gen_power`/`gen_power_chebyshev`. However, the helper is `pub(crate)` and fragile: `split_degree(0)` can underflow at `n - 1`, and `split_degree(1)` returns `(0, 1)`.

Suggestion: add `debug_assert!(n >= 2, "split_degree expects n >= 2")`, or make it return a `Result` if broader internal use is expected.

poulpy-ckks/src/api/polynomial_evaluation.rs
```diff
@@ -0,0 +1,92 @@
...
+    Even,
+    /// Odd polynomial: only odd-degree coefficients are non-zero.
+    Odd,
+}
+
+pub trait BSGSPolynomialInfos<BE: Backend> {
+    type Coeffs: GLWEToBackendRef<BE> + CKKSCtBounds;
+    fn degree(&self) -> usize;
+    fn baby_steps(&self) -> usize;
+    fn baby_step(&self, i: usize) -> &Self::Coeffs;
```
Janmajayamall: Low: `BSGSPolynomialInfos::baby_step(i)` can panic by contract, but API does not document it

The concrete implementation indexes directly into `baby_steps`, so invalid `i` will panic. This is likely fine for the internal evaluator, but the public trait should document that callers must only call `baby_step(i)` for `0 <= i < baby_steps()`.

Suggestion: add method-level docs, or return `Option<&Self::Coeffs>` if this trait is intended for safer external use.

poulpy-ckks/src/api/polynomial_evaluation.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn get(&self) -> &Self::Value;
+    fn get_mut(&mut self) -> &mut Self::Value;
+}
+
+pub trait PowerBasisHelper<BE: Backend, A> {
+    fn get(&self, power: usize) -> Result<&A>;
+}
+
+pub trait PolynomialEvaluation<BE: Backend> {
+    fn ckks_eval_baby_step<R, C, A, G>(
```
Janmajayamall: Low: `ckks_eval_baby_step` takes `parity` separately from `coeffs`

This is flexible, but it allows callers to pass a parity that does not match the encoded coefficient plaintext. For example, passing `Parity::Even` for coefficient plaintexts containing odd terms would silently skip those terms.

Suggestion: document the invariant clearly: `parity` must match the encoded coefficient plaintext.

poulpy-ckks/src/default/polynomial_evaluation.rs
```diff
@@ -0,0 +1,331 @@
...
+        R: GLWEToBackendMut<BE> + CKKSCtBounds + SetCKKSInfos,
+        B: BSGSPolynomialInfos<BE>,
+        A: GLWEToBackendRef<BE> + CKKSCtBounds,
+        G: PowerBasisHelper<BE, A>,
+        T: GGLWEInfos + GLWETensorKeyPreparedToBackendRef<BE>,
+        CKKSCiphertext<BE::OwnedBuf>: GLWEToBackendMut<BE> + GLWEToBackendRef<BE>,
+        for<'b> ScratchArena<'b, BE>: ScratchAvailable + ScratchArenaTakeCore<'b, BE>,
+    {
+        ensure!(
+            poly.baby_steps() > 0,
```
Janmajayamall, line 194: Medium: full evaluator should validate polynomial/power-basis basis compatibility

The evaluator has both `poly` and `power_basis` here, but it does not validate that they use the same basis. Later `power_basis.get(i)` calls are basis-dependent: for monomial they mean `X^i`, while for Chebyshev they mean `T_i(X)`. A mismatch can silently return the wrong result if the required indexes exist.

Suggestion: add `basis()` to `PowerBasisHelper` and check `poly.basis() == power_basis.basis()` once near the top of this function.

poulpy-ckks/src/default/polynomial_evaluation.rs
```diff
@@ -0,0 +1,331 @@
...
+            "ckks_eval_poly_real_const_coeffs_from_power_basis: polynomial must contain at least one baby step"
+        );
+
+        let n_baby = poly.baby_steps();
+        let last_coeffs = poly.baby_step(n_baby - 1);
+        let trailing_const_only = n_baby >= 2 && last_coeffs.n().as_usize() == 1;
+        let fold_power = poly.degree();
+        let can_fold = trailing_const_only && power_basis.get(fold_power).is_ok();
```
Janmajayamall: Low: `can_fold` swallows all `power_basis.get(fold_power)` errors

With the current `PowerBasis`, an error mostly means “missing power”, so falling back is reasonable. However, `PowerBasisHelper::get` returns a generic `Result`; custom implementations could return errors for other reasons. This line would silently discard those errors and continue down another evaluation path.

Suggestion: add a separate `contains(power)`/`has_power(power)` method to `PowerBasisHelper`, or document that `get()` errors must only mean “missing power”.

poulpy-ckks/src/test_suite/polynomial_evaluation.rs
```diff
@@ -0,0 +1,746 @@
...
+    E: NegacyclicFFT<F> + NegacyclicFFTNew<F>,
+{
+    let m = params.n / 2;
+    let encoder = Encoder::<E>::new(m).unwrap();
+    let (x_re_raw, _x_im_raw) = test_vector_1::<F>(m);
+    let x_im_raw = vec![F::zero(); x_re_raw.len()];
+    let input_meta = precision_at(&params, params.prec.log_delta.min(20));
+    let (x_re, _) = quantized_slots(host_module, &encoder, params.base2k.into(), input_meta, &x_re_raw, &x_im_raw);
+
+    let poly = Polynomial::chebyshev_interpolate(31, -F::one(), F::one(), |x: F| x.sin())
```
Janmajayamall, line 764: Medium: Chebyshev degree-31 expected value mirrors the encoded BSGS representation, not the original polynomial

`eval_encoded_bsgs_chebyshev` evaluates the already-decomposed/encoded BSGS structure using the same giant-step shape as the homomorphic evaluator. This checks that encrypted evaluation matches the encoded BSGS representation, but it does not independently verify that `encode_bsgs` preserved the original Chebyshev polynomial.

Suggestion: also compare against an independent reference, for example `poly.evaluate(x)` or directly against `x.sin()` within interpolation tolerance.

poulpy-ckks/src/test_suite/polynomial_evaluation.rs
```diff
@@ -0,0 +1,746 @@
...
+        .collect();
+    let want_re: Vec<F> = x_re_raw
+        .iter()
+        .map(|&x| encoded_coeffs.iter().rev().fold(F::zero(), |acc, &c| acc * x + c))
+        .collect();
+    let want_im = vec![F::zero(); x_re_raw.len()];
+
+    let poly_ref = Polynomial::new(Basis::Monomial, raw_coeffs.to_vec());
+    let bsgs_host = poly_ref
+        .encode_bsgs(host_module, params.base2k.into(), PT_PREC)
```
Janmajayamall, line 515: Medium: no integration test for `SplitStrategy::MinMult`

Most tests call `encode_bsgs(...)`, which uses the default `SplitStrategy::MinDepth`. There are unit tests in `polynomial.rs` for some decomposition shapes, but no encrypted/evaluation test for `encode_bsgs_with(..., SplitStrategy::MinMult)`. This leaves `MinMult` without end-to-end coverage through encoding, power-basis population, and CKKS evaluation.

Suggestion: add at least one monomial and/or Chebyshev evaluation test using `SplitStrategy::MinMult`.

poulpy-ckks/src/default/plaintext.rs
```diff
@@ -15,21 +16,45 @@ use crate::{
     ensure_plaintext_degree_match,
 };
 
+#[derive(Clone, Copy)]
+enum PlaintextShift {
+    Rsh(usize),
+    Lsh(usize),
+}
+
+fn plaintext_shift<P: CKKSInfos + LWEInfos>(ct_log_budget: usize, pt: &P) -> PlaintextShift {
```
Janmajayamall, line 25: Medium: `plaintext_shift` uses `pt.max_k()` while alignment validation uses `pt.effective_k()` (low confidence)

Callers validate alignment using `pt.effective_k()`, but the actual shift is computed using `pt.max_k()`. That means validation is based on the plaintext semantic width while shifting is based on allocated/rounded storage width. This may be intentional if encoded coefficients are expected to occupy the top of the allocated limb span; if so, it should be documented clearly. If not, `plaintext_shift` likely should use `pt.effective_k()` instead.

This affects the plaintext add/sub paths that validate alignment before calling this helper.

poulpy-ckks/src/layouts/ciphertext.rs
```diff
@@ -104,6 +106,22 @@ impl<D: Data, S: CKKSNormalizationState> CKKSCiphertext<D, S> {
         self.meta = meta;
         Ok(())
     }
+
+    /// Allocates a fresh backend-owned ciphertext and copies `self` into it.
+    ///
+    /// Used to normalize the allocation size after arithmetic operations that
+    /// may leave the buffer over-sized relative to `effective_k`.
+    pub fn compact<M, BE>(&self, module: &M, scratch: &mut ScratchArena<'_, BE>) -> Result<CKKSCiphertext<BE::OwnedBuf>>
```
Janmajayamall, line 116: Medium: `compact` is implemented for both normalized and unnormalized ciphertexts, but always returns normalized typestate

Because this method is implemented on `CKKSCiphertext<D, S>`, it is callable for both normalized and unnormalized ciphertexts. However, it always returns `CKKSCiphertext<BE::OwnedBuf>`, which defaults to the normalized typestate. `ckks_copy` copies/shifts limbs but does not carry-normalize unnormalized data, so `compact()` on an `Unnormalized` ciphertext could incorrectly mark unnormalized data as normalized.

Suggestion: restrict `compact` to `CKKSCiphertext<D, Normalized>`, or preserve the normalization typestate in the return type if feasible.

poulpy-cpu-ref/examples/ckks_poly2.rs
```diff
@@ -377,172 +271,114 @@ fn encryption(setup: &mut SetupArtifacts, encoding: &EncodingArtifacts) -> Resul
...
-            let left_im = encoding.coeffs.a.1 + encoding.coeffs.b.0 * xi + encoding.coeffs.b.1 * xr;
-            let right_re = encoding.coeffs.c.0 + encoding.coeffs.d.0 * xr - encoding.coeffs.d.1 * xi;
-            let right_im = encoding.coeffs.c.1 + encoding.coeffs.d.0 * xi + encoding.coeffs.d.1 * xr;
-            left_im + right_re * x2i + right_im * x2r
-        })
-        .collect();
-
-    let err_re = max_err(&decryption.have_re, &want_re);
-    let err_im = max_err(&decryption.have_im, &want_im);
+    let avg_err_re = avg_err(&decryption.have_re, &want_re).log2();
```
Janmajayamall: Low: verification error naming/threshold is confusing

The variables store `log2(error)`, but the printed labels use `|have − sin(x)|`, which reads like absolute error. The threshold `5e-5` also looks like an absolute tolerance, not a log2 threshold. If this is intended to be log2 error, the labels and threshold should make that explicit. If it is intended to be absolute error, the `.log2()` calls should likely be removed.

Question: is `5e-5` meant to be an absolute error tolerance or a log2-error threshold?

poulpy-cpu-ref/examples/ckks_poly2.rs
```diff
@@ -34,30 +38,38 @@ use poulpy_core::{
...
 type SecretKeyPrepared = GLWESecretPrepared<<BackendImpl as Backend>::OwnedBuf, BackendImpl>;
 type TensorKeyPrepared = GLWETensorKeyPrepared<<BackendImpl as Backend>::OwnedBuf, BackendImpl>;
 
-const N: usize = 4096;
+const DEGREE: usize = 31;
+const N: usize = 1024;
 const M: usize = N / 2;
 const BASE2K: usize = 52;
-const CT_K: usize = 95;
+/// Seven limbs of 52 bits = 364 bits total capacity.
```
Janmajayamall: Low: `CT_K` comment may be stale

This comment looks potentially stale: `CT_K` is `300`, `PREC_CT.log_delta` is `45`, and `DEGREE` is `31`, while the comment refers to seven 52-bit limbs, log_delta `51`, and degree `63`.

Question: is this comment stale, or are these numbers still intentional?

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+- Generalize fused CKKS multiply-add/plaintext paths over backend-owned ciphertext/plaintext layouts and allow plaintext add/sub alignment to left-shift when the encoded plaintext precision exceeds the ciphertext budget, enabling coefficient-indexed plaintext constants used by polynomial evaluation.
+
+### `poulpy-hal` / `poulpy-cpu-ref`
+- Add `VecZnxLshAddCoeffToCoeffBackend` and `VecZnxLshSubCoeffToCoeffBackend` hooks plus portable reference implementations so coefficient-level plaintext accumulation can handle left-shift alignment.
+
+### `poulpy-bench`
+- Add the `ckks_poly_eval` Criterion benchmark, sweeping polynomial degree and `MinDepth` / `MinMult` BSGS split strategies on `ntt120-ref` while reporting baby-step size and observed log-budget/level consumption.
+
+### Build & Docs
+- Add `docs/polynomial-evaluation-spec.md`, a scheme-agnostic design note for power bases, Paterson-Stockmeyer/BSGS decomposition, encoded baby polynomials, and the Poulpy CKKS integration.
```
Janmajayamall: Low: changelog references a docs file that is not present in the PR

`docs/polynomial-evaluation-spec.md` is not present in the PR diff and does not exist in the checkout. This looks stale, especially since the PR includes a commit titled “removed outdated poly-eval spec”.

Suggestion: remove or update this changelog entry.

poulpy-ckks/src/polynomial.rs
Pro7ech: Intended behavior, false positive.

poulpy-ckks/src/polynomial.rs
Pro7ech: Also intended behaviour. If parity could be automatically derived, there wouldn't be a need to pass it manually.

poulpy-ckks/src/polynomial.rs
Pro7ech: False positive -> there is a direct API for constant multiplication.

poulpy-ckks/src/polynomial.rs
Pro7ech: This is also intentional.

poulpy-ckks/src/api/polynomial_evaluation.rs
Pro7ech: This is self explanatory.

poulpy-ckks/src/power_basis.rs
Pro7ech: duplicate

poulpy-ckks/src/default/polynomial_evaluation.rs
Pro7ech: duplicate

poulpy-ckks/src/default/plaintext.rs
Pro7ech: false positive

jdumezy (approved):

PR #193, opened 2026-08-06
Jp/word type

### LINKED ISSUE #162: Generalize VecZnx backend view
Allow for i32 or i128 VecZnx. Ensure base2k first within the coeff size.

Pro7ech:
### Summary

This PR introduces explicit word-typed HAL layouts so each polynomial container states which scalar word convention backs its storage. The main motivation is to prepare the HAL for future backends that use `i32` coefficient-domain words for non-prepared structs, while making today’s `i64`/`i128`/DFT/CRT word choices visible in the type system.

### Highlights

- Added explicit word traits: `ZnxWord`, `BigWord`, and `DftWord`.
- Added a defaulted word parameter to non-prepared layouts like `VecZnx`, `ScalarZnx`, and `MatZnx`, preserving existing `i64` call sites while opening the path to `i32`.
- Re-keyed prepared and big-domain containers by word type, so their storage convention is explicit at the type level.
- Added prime-set-parameterized `CrtWord<P, T>` for CRT/NTT prepared words, making prime set, lane element, and lane geometry part of the type.
- Added backend-keying and layout-compatibility markers for prepared/big containers, so zero-copy cross-backend reuse is explicit and checked.
- Kept the current coefficient-domain op layer fenced to `Backend<ZnxWord = i64>` until the future `i32` backend plumbing lands.


jdumezy:
Few findings (?) I think could be worth fixing:
- View validation seems still unsound for malformed metadata: unchecked size/offset arithmetic can wrap before `from_raw_parts`, and alignment is only checked in debug builds.
- `impl_backend_from!` does not forward sizing overrides or SCRATCH_ALIGN, yet emits all unsafe compatibility markers. An IFMA delegate, for example, computes 24 instead of 16 bytes per coefficient.
- Derived Eq appears to still requires W: Eq through `PhantomData`, so FFT containers using f64 do not implement Eq, contrary to the changelog.
- The changelog says the local AArch64/NEON lane is documented in `ci-local.md`, but that file is absent

jdumezy (approved):

PR #197, opened 2026-08-07
Jdmz/libm prep

jdumezy:
Add approximation of functions for CKKS


poulpy-ckks/src/approximation/remez.rs
```diff
@@ -0,0 +1,705 @@
+//! Host-side Remez fitting in the Chebyshev basis.
```
Pro7ech: Could this file be fully documented? Like reference to each algo, how it works, what it does, where it fits at a high level in the steps of the Remez algo, etc... Maybe even split this file into multiple different files.

Pro7ech (approved):

PR #198, opened 2026-08-07
replace many vec<u8> and i64 by generic BE::OwnedBuff and BE::ZnxWord

Pro7ech:
What the title says (moving toward making it possible to add 32bit backends). Add `BE::ZnxWord` generic to `VecZnx` and others.


jdumezy:
Overall lgtm, with 3 potential caveats:
- It looks like cross-backend coefficients transfers copy raw bytes and retag them as `To::ZnxWord` without requiring word/layout compatibility, so differing word sizes or padding can silently corrupt data
- `MatZnx` allocates using backend-overridable sizing but computes backend entry strides using the dense host layout, so padded backends seem to address the wrong entries
- `lwe_bytes_of_from_infos` sums body + mask sizes without the alignment padding

jdumezy (approved):

Issue #163, opened 2026-05-18
Rename secret distribution metadata from dist to base_dist


Janmajayamall:
The `dist` field on secret key structs is currently used to track the distribution of the **base/original sampled secret**, not necessarily the current coefficient distribution after transformations such as automorphisms or conversions.

This can be confusing: `dist` sounds like it describes the current coefficient distribution, while in practice it is metadata about the base sampling distribution.

Example confusion:

In `poulpy-core/src/layouts/glwe_secret.rs`, `lwe_secret_from_glwe_secret` does:

```rust
res.dist = src.dist;

let src_vec = scalar_znx_as_vec_znx_backend_ref_from_ref::<B>(&src.data);
let res_as_backend = <ScalarZnx<B::OwnedBuf> as ScalarZnxToBackendMut<B>>::to_backend_mut(&mut res.data);
let mut res_vec = VecZnx::from_data(res_as_backend.data, n, rank, 1);

for j in 0..rank {
    self.vec_znx_automorphism_backend(-1, &mut res_vec, j, &src_vec, j);
}
```

At first glance, this looks suspicious because `dist` is copied unchanged even though coefficients are transformed by an automorphism. If `dist` is interpreted as “current coefficient distribution,” this appears incorrect.

But if `dist` means “base/original sampled secret distribution,” then copying it is intentional. Renaming it to `base_dist` would make that intent clear.

Proposal:
- Rename `dist` to `base_dist`
- Rename related accessors if applicable:
  - `dist()` → `base_dist()`
  - `dist_mut()` → `base_dist_mut()`
- Update docs to clarify that this metadata describes the original/base secret distribution used for security/noise assumptions, not necessarily the current transformed coefficient support.

Pro7ech:
#206

PR #206, opened 2026-08-11
Jp/secrets fixes

### LINKED ISSUE #141: User shouldn't be able to give GLWESecretKey to prepare of GLWESecretTensorPrepared
Prepare of GLWESecretTensorPrepared should only accept a GLWESecretTensor

Pro7ech:
- fix #141 and #163
- Move secret sampling behind HAL API, forcing backends to implement it
- Fix latent bug in glwe-tensoring noise assessment


jdumezy:
Few comments:
- A nonzero SHIP secret is tagged as ZERO
- It might be good to update the doc in `poulpy-ckks/src/layouts/bootstrapping_keys.rs`

jdumezy (approved):

PR #207, opened 2026-08-12
Jp/qol

Pro7ech:
- Move SlotKind to CKKSMeta
- Update PACO to auto-derive calls based on LogSparsity
- Condense bootsrapping API with auto-dispatches based on SlotKind
- Retroactively condense the changelog


jdumezy:
Overall lgtm, with two issues:
- It looks like imaginary constant additions can remain marked Real
- Also, C2S-first bootstrapping loses real-slot metadata (contrary to S2C-first bootstrapping)

jdumezy (approved):

PR #209, opened 2026-08-12
update noise helper

### LINKED ISSUE #204: Update noise equation to use k_aux
`k_aux` was added, which greatly simplified switching key parameterization. This simplification must be extended to the noise equations used in the tests.

Pro7ech:
fixes #204


jdumezy:
- GGSW encryption tests seem to still partially use the pre-k_aux precision, and build the bound from the synthetic k, unlike the other converted tests (`poulpy-core/src/test_suite/encryption/ggsw_ct.rs`)
- I think source and destination secret variances are sometimes conflated in `poulpy-core/src/default/noise/mod.rs` line `98`. The same `var_xs` is used for key-mask error. This is also probably problematic for sparse key encapsulation.
- In the same file line `218`, the keyswitch residue equation is reused. Couldn't this be a problem for the external product, as dropped body and mask componennts are multiplied by the GGSW message and require `var_msg` weighting. Currently even `varèmsg == 0` produces a nonzero residue contribution, and other message distributions couls be under/overestimated?
- Same file again where `rank_in` is changed from `rank` to `rank+1`. This seems to inflate the tsk terms by (rank + 1)/rank.

jdumezy (approved):

PR #210, opened 2026-08-12
Restructure `poulpy-bench` as a generic benchmarking API instantiated by the backend crates

### LINKED ISSUE #124: Add benchmarks for core operations

ChristianMct:
## Summary

Reworks `poulpy-bench` into a layered, composable framework that any backend crate can
import to instantiate its benchmarks without writing its own benchmark logic. The crate
defines a generic runner for each backend operation, and a generic sweep driver that turns
a table of runners plus a param sweep into Criterion groups. Backends can compose their
own benchmark suites by selecting the operations they implement and the sweeps they want
to run. 

## Motivation

`poulpy-bench` owned its own `[[bench]]` binaries and depended on every backend crate  to
drive them via feature-gated `for_each_backend!` macros. This required an update on
`poulpy-bench` to benchmark a new backend and was also not very ergonomic. For example, a
partially implemented backend would have needed a custom treatment by `poulpy-bench` to
avoid running benchmarks for the operations it did not implement. It should be the other
way around: a backend crate should be able to compose its own benchmark suite by selecting
the operations it implements and the sweeps it wants to run.

### `poulpy-bench` API

- A **runner** is `fn<B, M>(&mut Bencher<'_, M>, &P)` that benchmarks a specific operation
  of a backend `B` over a set of parameters `P`: it allocates its inputs once, then times
  the operation via `bencher.iter(...)`. Each runner's `where` clause sets a backend trait
  bound on `B` that is minimal for the operation it benchmarks.
- **`BenchOp<M, P>`** is a runner associated with a name.
- A benchmark **suite** is a table of `BenchOp<M, P>`, and the crate provides a generic
  `bench_ops` function that takes a table of runners plus a param sweep and turns it into
  Criterion groups.
- All of these types and functions are `pub`, so a backend crate can build fully custom
  suites. 

###  For all backend crates:

Three Criterion targets built from the API above:

- `benches/full.rs`: full sweep across HAL, core, and CKKS layers.
- `benches/standard.rs`: subset of the most relevant operations at each layers.
- `benches/light.rs`: a fast benchmark for sanity checks, the standard suite over a single parameter set for `log_n = 1 << 14`.

### Notes

- This PR removes the per-layer features from `poulpy-bench`, because they are no longer required to control which benchmarks are run and  `poulpy-bench` can now be a dev-dependency of the backend crates. But this might be an assumption on my side of how these features where used.
- This PR fixes a bug in the `glwe_mul_plain_tmp_bytes` function implementation. It is the only change made to pre-existing code outside of the `poulpy-bench` crate.

## Current limitations

- There are currently no benchmarks for the composite operations (polyeval, lintrans,
  evalmod, BDD arithmetic). The not-yet-adapted benchmarks are currently in
  `poulpy-bench/old_bench` (not built).
- Not all CKKS operations have been mapped to runners yet. However most of the skipped operations
  are variants such as `*_assign` ops.
- The param sweeps are quite arbitrary and need to be reviewed and set to more relevant
  values. This is due to my relative unfamiliratity with the bivariate representation.
- I still need to check the recent changes on main to rebase.

Tasks:

- [x] Pass CI tests
- [x] Update the README
- [x] Decide on a light benchmark suite and params
- [x] Add benches to all backends
- [x] Decide if benchmarks should be tested by CI (Criterion's `--test` flag)


poulpy-bench/src/core/external_product.rs
```diff
@@ -32,28 +32,43 @@ pub fn bench_glwe_external_product<BE: Backend<OwnedBuf = Vec<u8>, ZnxWord = i64
...
+    let mut ct_ggsw: GGSW<Vec<u8>, i64> = module.ggsw_alloc_from_infos(&ggsw_infos);
+    let mut ct_glwe_in: GLWE<Vec<u8>, i64> = module.glwe_alloc_from_infos(&glwe_infos);
+    let mut ct_glwe_out: GLWE<Vec<u8>, i64> = module.glwe_alloc_from_infos(&glwe_infos);
 
     let mut scratch: ScratchOwned<BE> = ScratchOwned::alloc(
-        module.ggsw_encrypt_sk_tmp_bytes(ggsw_infos)
-            | module.glwe_encrypt_sk_tmp_bytes(glwe_infos)
-            | module.glwe_external_product_tmp_bytes(glwe_infos, glwe_infos, ggsw_infos),
+        module.ggsw_encrypt_sk_tmp_bytes(&ggsw_infos)
+            | module.glwe_encrypt_sk_tmp_bytes(&glwe_infos)
```
Pro7ech: Exclusive `or` should be replaced with `max` (everywhere)

poulpy-bench/src/core/operations.rs
```diff
@@ -0,0 +1,182 @@
...
+    ScratchOwned<BE>: ScratchOwnedAlloc<BE> + ScratchOwnedBorrow<BE>,
+    for<'x> BE::BufMut<'x>: HostDataMut + AsRef<[u8]> + AsMut<[u8]> + Sync,
+{
+    let infos = glwe_layout(cp);
+    let module: Module<BE> = Module::<BE>::new(cp.n as u64);
+
+    let mut ct_out: GLWE<Vec<u8>, i64> = module.glwe_alloc_from_infos(&infos);
+    let ct_in: GLWE<Vec<u8>, i64> = module.glwe_alloc_from_infos(&infos);
+    let pt: GLWEPlaintext<Vec<u8>, i64> = module.glwe_plaintext_alloc_from_infos(&infos);
+    let mut scratch: ScratchOwned<BE> = ScratchOwned::alloc(module.glwe_mul_plain_tmp_bytes(&ct_out, &ct_in, &pt) * 2); // TODO: have to double it for some reason.
```
Pro7ech: `glwe_mul_plain_tmp_bytes_default` should be updated with

```rust
let res_dft_size = a_size + b_size;
let lvl_2_cnv_apply: usize = self.cnv_apply_dft_lazy_tmp_bytes(0, res_dft_size, a_size, b_size);
```

poulpy-bench/src/core/suites.rs
```diff
@@ -0,0 +1,344 @@
...
+        + GLWESwitchingKeyEncryptSk<BE>
+        + GLWEKeyswitch<BE>
+        + GLWESwitchingKeyPreparedFactory<BE>
+        + GLWETensoring<BE>
+        + GLWETensorKeyPreparedFactory<BE>
+        + Convolution<BE>
+        + CnvPVecAlloc<BE>
+        + VecZnxIdftApplyTmpA<BE>
+        + VecZnxBigNormalize<BE>
+        + VecZnxSubAssignBackend<BE>
```
Pro7ech: Double check why HAL bounds leak on poulpy-core bounds.

poulpy-bench/src/hal/vec_znx/rotate.rs
```diff
@@ -0,0 +1,63 @@
...
+    Module<B>: VecZnxRotateBackend<B> + ModuleNew<B>,
+    B::OwnedBuf: AsMut<[u8]>,
+{
+    let module: Module<B> = Module::<B>::new(sweep.n as u64);
+
+    let mut source: Source = Source::new([0u8; 32]);
+
+    let mut a = module.vec_znx_alloc(sweep.cols, sweep.size);
+    let mut res = module.vec_znx_alloc(sweep.cols, sweep.size);
+    source.fill_bytes(a.data_mut().as_mut());
```
Pro7ech: Make sure source byte filling is replace everywhere by ` let a = random_host_vec_znx(module.n(), sweep.cols, sweep.size, &mut source);`

poulpy-bench/src/hal/vec_znx/sub.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let mut a = module.vec_znx_alloc(sweep.cols, sweep.size);
+    let mut b = module.vec_znx_alloc(sweep.cols, sweep.size);
+    let mut c = module.vec_znx_alloc(sweep.cols, sweep.size);
+    source.fill_bytes(a.data_mut().as_mut());
+    source.fill_bytes(b.data_mut().as_mut());
+    source.fill_bytes(c.data_mut().as_mut());
+
+    bencher.iter(|| {
+        let a = <VecZnx<B::OwnedBuf, B::ZnxWord> as VecZnxToBackendRef<B>>::to_backend_ref(&a);
```
Pro7ech: I believe this full declaration is an artifact and can be removed. If yes, double check everywhere.

poulpy-bench/src/hal/helpers.rs
```diff
@@ -0,0 +1,208 @@
...
+#[allow(dead_code)]
+pub fn vec_znx_big_backend_ref<'a, BE: Backend<ZnxWord = i64>>(src: &'a VecZnxBigOwned<BE>) -> VecZnxBigBackendRef<'a, BE> {
+    src.to_backend_ref()
+}
+
+pub fn vec_znx_big_backend_mut<'a, BE: Backend<ZnxWord = i64>>(src: &'a mut VecZnxBigOwned<BE>) -> VecZnxBigBackendMut<'a, BE> {
+    src.to_backend_mut()
+}
+
+// pub fn upload_host_glwe<BE>(module: &Module<BE>, src: &GLWE<Vec<u8>, i64>) -> GLWE<BE::OwnedBuf, BE::ZnxWord>
```
Pro7ech: remove commented code

poulpy-bench/src/hal/helpers.rs
```diff
@@ -0,0 +1,208 @@
...
+    cols: usize,
+    size: usize,
+    source: &mut Source,
+) -> CnvPVecROwned<BE> {
+    let mut bytes = vec![0u8; BE::bytes_of_cnv_pvec_right(n, cols, size)];
+    source.fill_bytes(&mut bytes);
+    CnvPVecROwned::<BE>::from_bytes(n, cols, size, bytes)
+}
+
+pub fn scalar_znx_backend_ref<'a, BE: Backend<ZnxWord = i64>>(
```
Pro7ech, line 125: Are these helpers still needed?

poulpy-bench/src/params.rs
```diff
@@ -1,85 +1,61 @@
...
 
-/// HAL sweep parameters for `vec_znx*`, `vec_znx_dft`, and `svp` benchmarks.
-///
-/// Each entry is `[log_n, cols, size]`.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(default)]
-pub struct HalSweepParams {
-    pub sweeps: Vec<[usize; 3]>,
+#[derive(Debug, Clone)]
+pub struct HalSweepParms {
```
Pro7ech: I'd put the params in the mod.rs of their respective module

poulpy-cpu-ref/benches/full.rs
```diff
@@ -0,0 +1,140 @@
+//! Full benchmark binary: every op family across HAL, core, CKKS, and
```
Pro7ech, line 1: Ce serait mieux de séparer les layers plutôt que de toutes les avoir dans le même benchmark

poulpy-bench/src/core/suites.rs
ChristianMct: The benchmark runners which are leaking these bounds:
- `runner_glwe_tensor_prepare_left` and `runner_glwe_tensor_prepare_right` are benchmarking `Convolution<BE>.cnv_prepare_left` and `Convolution<BE>.cnv_prepare_right`. So they seem to duplicate the HAL benchmarks (except with different sweep params).
- `runner_glwe_tensor_diag_lane` and `runner_glwe_tensor_pairwise_lane` are benchmarking a composition of several module operations. My understanding is that they are "sub-benchmarks" of parts of `glwe_tensor_apply`.

poulpy-bench/src/hal/helpers.rs
ChristianMct: For now, yes. The module method expect specific type that is not the one returned by the `upload_*` functions. One option could be to fix these upload to return the right type directly ? I don't fully get the design of the upload+reference workflow so I left it as is.

poulpy-cpu-ref/benches/full.rs
ChristianMct: This would require one file for each [light, standard, full] X [HAL, Core, Schemes]. Alternative could be to have the layer as a prefix to the benchmark label (I added that in my last commit). So one can select which layer is run with 

```
cargo bench -p poulpy-cpu-ref --bench full --features enable-core,enable-ckks -- "/hal/"
```

poulpy-bench/src/core/external_product.rs
ChristianMct: Done.

poulpy-bench/src/core/operations.rs
ChristianMct: Done, note that this is the only change outside of `poulpy-bench` and the backends' benchmarks.

poulpy-bench/src/hal/vec_znx/rotate.rs
ChristianMct: Done.

poulpy-bench/src/hal/vec_znx/sub.rs
ChristianMct: Have been replaced by the helper function call.

poulpy-bench/src/hal/helpers.rs
ChristianMct: Done.

poulpy-bench/src/params.rs
ChristianMct: Done.

poulpy-bench/src/core/suites.rs
Pro7ech: `runner_glwe_tensor_prepare_left` and `runner_glwe_tensor_prepare_right` are not part of the poulpy-core API, and `cnv_prepare_right` and `cnv_prepare_left` are part of the poulpy-hal API. If they are already benchmarked there, you can simply remove this benchmark. Else move it under poulpy-hal and rename it.
 
 Same for `runner_glwe_tensor_pairwise_lane`, it's not benchmarking a method of the poulpy-core API but a composition of multiple poulpy-hal ops, which should be already benchmarked there.

Pro7ech: More generally, a bench suite should only benchmark its exposed user API, so no bounds should leak outside of the crate it is benchmarking (except maybe for schemes where part of poulpy-core might be shared accross them).

poulpy-cpu-ref/benches/full.rs
Pro7ech: If carefully naming the labels solve this, that's fine too. What is most important is being able to target families or specific functions in a family.

ChristianMct: I streamlined the way benchmark render their IDs from the backend name and layers. Now `BenchOp` are directly tagged with their layers. I also documented in `poulpy-bench/README.md` how to use Criterion's filter to control which benchmarks are run.

poulpy-bench/src/core/suites.rs
ChristianMct: I removed the duplicates and composite benchmarks from the core suites, which removes the leaked bounds.

ChristianMct:
Currently, the CI does not ensure the benchmarks are building and running without error. This was already the case before this PR, and some benchmarks were actually resulting in runtime errors. 

Criterion has a `--test` flag that runs each benchmark for one iteration. But this would still make the CI checks longer. On my machine, for the current choice of parameter grid:
- `full` suite -> 224s 
- `standard` suite -> 44s

Pro7ech (dismissed):

Pro7ech (dismissed):

Pro7ech (approved):

PR #216, opened 2026-08-18
Make library more friendly to device backends

Pro7ech:
# Make the library more friendly to device backends

Everything below `poulpy-hal` already worked on a device backend; everything at
`poulpy-core` and above assumed its buffers were host memory it could
dereference. This closes that gap. 

> Most of it is visibility and bounds on abstractions that were already correct, not new design.

## New

**`poulpy-core`**

- `api::TransferInto`: one trait, implemented per layout, for moving a value
  between backends. Writes into a destination the caller has allocated:
  ```rust
  let mut sk = module.glwe_secret_alloc_from_infos(&sk_host);
  sk_host.transfer_into(&mut sk);
  ```
  No backend is named at a call site, and the whole shape is checked rather
  than just the byte count.
- `test_suite::parity` and `core_parity_test_suite!`: a cross-backend suite
  that runs one operation on a reference backend and on a backend under test
  over identical uniform inputs, and asserts byte equality. Covers key-switch
  (GLWE, assign, GGLWE), automorphism, external product, and the keyless GLWE
  operations. Involves no secrets, encryption or noise model, so a device
  backend can run it: it owes allocation, the prepared-key factory, the
  operation and transfer. 
  > Parity can only be tested on full E2E pipelines, i.e. inputs and outputs MUST be ScalarZnx, VecZnx or MatZnx.
- `GGLWEPrepared::data()`, generic over the buffer, so an out-of-tree backend
  can reach a prepared key's `VmpPMat` from the key-switch extension point.
- `BackendGLWESwitchingKey<BE>` / `BackendGLWEAutomorphismKey<BE>`.

**`poulpy-hal`**

- `CopyToHost` / `CopyFromHost` and the free `transfer_buf_into(src, dst)`.
  Keyed on the buffer rather than the backend, which is what removes every
  backend type parameter from the transfer surface. One copy whenever either
  side is host-visible; only device-to-device stages.
- `Backend::len_bytes_ref` / `len_bytes_mut`.
- `HalModuleImpl::Config` (defaulting to `()`) and `new_with(n, config)`,
  mirrored on `ModuleNew`, for device selection at construction.
- `layouts::vec_znx_backend_ref` / `vec_znx_backend_mut` /
  `vec_znx_reborrow_backend_mut`, re-exported by `test_suite`.

**`poulpy-bench`**

- `core::fill`: backend-native operand construction shared by the core runners.

**Backends**

- `poulpy-cpu-avx`, `-avx512` and `-arm` each run the parity suite against
  their `poulpy-cpu-ref` sibling, for the FFT64 and NTT4x30 families.

## Changed

**`poulpy-core`**

- `test_suite` splits into `test_suite::noise` (the existing scheme-correctness
  suite, moved wholesale) and `test_suite::parity`. `core_backend_test_suite!`
  is unchanged.
- Relaxed to `D: Data`, having been host-bounded for no reason:
  `GGLWE` / `GLWEPlaintext` / `LWEPlaintext` / `GLWETensor` `data()` /
  `data_mut()`, `GLWESwitchingKeyDegrees(Mut)`, `Get`/`SetGaloisElement`,
  `GetDistribution(Mut)`, `SetBase2k`.
- `GGLWEAtBackendRef`/`Mut` and `GGSWAtBackendRef`/`Mut` are public, and
  `GLWESwitchingKey`, `GLWEAutomorphismKey` and `GLWETensorKey` delegate them.
- `GLWEKeyswitchInternal` and `GGLWEProductDefault` are public; they appear in
  the bounds of the public `glwe_keyswitch*_default` functions.
- The `*Default` override surfaces are no longer `#[doc(hidden)]`, and `oep`'s
  module docs describe the `*Impl` / `*Default` split with a worked override.
- The gadget-digit width rule moves from an in-function comment to the
  `GLWEKeyswitchDefault` contract.

**`poulpy-hal`**

- `HostStaged` becomes `Backend<ZnxWord = i64, OwnedBuf: CopyToHost + CopyFromHost>`.

**`poulpy-bench`**

- All 19 `core` runners drop `OwnedBuf = Vec<u8>` and the host-view bounds, so
  a device backend can run them. Operands that are not the subject of the
  measurement are uniform noise filled through the backend rather than
  encrypted.
- Fixed: the operation and tensor runners timed on freshly allocated, zeroed
  buffers, which can hit float-FFT denormals.

**`poulpy-cpu-ref`**

- Crate docs document the `enable-core` / `enable-ckks` features.

## Removed

- `ModuleTransfer` and the inherent `Layout::to_backend` methods — 29 methods
  across three spellings of one operation, replaced by `TransferInto`.
- `TransferFrom`, replaced by `CopyToHost` / `CopyFromHost`, along with the ~44
  hand-written per-backend-pair impls it required.
- `GGLWEPreparedVmpPMatRef`, superseded by `GGLWEPrepared::data()`.
- 47 dead `#[allow(private_bounds)]` from `oep`, and the remaining
  `#![allow(private_bounds)]` in `default` with them.
- Two dead `LookupTable::transfer_to` methods in `poulpy-bin-fhe`.


poulpy-core/src/oep/operations.rs
```diff
@@ -891,8 +876,8 @@ macro_rules! impl_glwe_rotate_impl_from {
                     <$be as poulpy_hal::layouts::Backend>::ZnxWord,
                 > = res_host.reinterpret::<$be>();
 
-                let a_delegate = $crate::api::ModuleTransfer::upload_glwe::<$be>(&delegate, &a_src);
-                let mut res_delegate = $crate::api::ModuleTransfer::upload_glwe::<$be>(&delegate, &res_src);
+                let a_delegate = $crate::api::TransferInto::upload_glwe::<$be>(&delegate, &a_src);
```
jdumezy: TransferInto only provides transfer_into. It has no upload_glwe or download_glwe methods. Any downstream expansion of this exported macro will fail to compile.

poulpy-bin-fhe/src/blind_rotation/lut.rs
```diff
@@ -242,21 +242,7 @@ impl<D: Data, W: ZnxWord> LookupTable<D, W> {
...
-        LookupTable {
-            data: self.data.iter().map(|glwe| glwe.to_backend::<From, To>(dst)).collect(),
-            rot_dir: self.rot_dir,
-            base2k: self.base2k,
-            k: self.k,
-            drift: self.drift,
-        }
-    }
-}
+impl<D: HostDataRef, W: ZnxWord> LookupTable<D, W> {}
```
jdumezy: Removing LookupTable::to_backend without adding TransferInto seems to leave device backends unable to construct or upload LUTs?

poulpy-bench/src/core/decryption.rs
```diff
@@ -16,14 +16,17 @@ use criterion::{Bencher, measurement::Measurement};
 
 use crate::core::params::CoreParams;
 
-pub fn runner_glwe_decrypt<BE: Backend<OwnedBuf = Vec<u8>, ZnxWord = i64> + HostBackend, M: Measurement>(
+pub fn runner_glwe_decrypt<BE: Backend<ZnxWord = i64> + HostBackend, M: Measurement>(
```
jdumezy: The HostBackend bound keeps this runner (and therefore decryption_ops, all_ops, and standard_ops) unusable by device backends, which could be an issue.

poulpy-hal/src/layouts/mod.rs
```diff
@@ -325,7 +333,7 @@ pub type OwnedBuf<BE> = <BE as Backend>::OwnedBuf;
 ///
 /// [`Self::transfer_buf`] receives an opaque buffer with no shape (`n`, `cols`,
```
jdumezy: Stale doc since Self::transfer_buf was removed

poulpy-core/src/oep/operations.rs
Pro7ech: I forgot to remove it, it has no use anymore.

poulpy-bin-fhe/src/blind_rotation/lut.rs
Pro7ech: Right, I'll add the `TransferInto` impl. It still cannot be constructed on devices because of the bounds on `LookupTableFactory`, some ops are done on `vec<i64>` directly. I think it's ok for now to expect them to be constructed on host and uploaded as this is, for bin-fhe, something that is only done once during setup.

poulpy-bench/src/core/decryption.rs
Pro7ech: Yes right, this bound isn't needed anymore, removing it.

poulpy-hal/src/layouts/mod.rs
Pro7ech: Right, I removed them, thank you

jdumezy (approved):

PR #226, opened 2026-08-21
Jp/poulpy bin fhe deps

Pro7ech:
### Summary

Remove backend deps in poulpy-bin-fhe. This meant removing examples, but we can add them later in a specialized crate compiling examples. And duplicating example in each backend is not a good practice.

### `poulpy-hal`

- Add `reference::znx`, the portable scalar kernels over `[i64]`, moved out of `poulpy-cpu-ref` so that a crate can reach them without depending on a backend. `poulpy_cpu_ref::reference::znx` re-exports them, so existing paths are unchanged.

### `poulpy-bin-fhe`

- **Breaking:** the crate no longer depends on any backend, in `dependencies` or `dev-dependencies`, matching `poulpy-core` and `poulpy-ckks`. The `enable-avx`, `enable-avx512f`, `enable-ifma` and `enable-neon` features are removed (`enable-neon` was inert: nothing referenced `poulpy-cpu-arm`), and `enable-bin-fhe` no longer enables anything.
- **Breaking:** the per-backend test modules are replaced by public `blind_rotation::test_suite`, `circuit_bootstrapping::test_suite` and `bdd_arithmetic::test_suite`, instantiated by a backend crate through the new `bin_fhe_backend_test_suite!`. The `bdd_arithmetic` / `circuit_bootstrapping` / `max_array` examples move to `poulpy-cpu-ref` and run on `FFT64Ref`.


jdumezy:
I think there is an issue with CI: poulpy-bin-fhe/enable-avx remains in AVX_FEATURES, while line 41/42 also retain enable-avx/enable-ifma.
Also, `cargo test -p poulpy-cpu-ref` fails because of a missing enable-core feature.
Finally, README in poulpy-bin-fhe, avx and avx512 are now stale with this change.

jdumezy (approved):

PR #229, opened 2026-08-25
Fix precision loss when encrypting with non limb-aligned k

jdumezy:


poulpy-ckks/src/test_suite/mul.rs
```diff
@@ -124,6 +133,84 @@ where
     );
 }
 
+pub fn test_mul_ct_limb_boundary_precision<BE, F, E>(
```
Pro7ech: Since the bug was encryption related, why do we need this test?

poulpy-core/src/default/encryption/compressed/gglwe.rs
```diff
@@ -169,6 +174,7 @@ where
                         source_xe,
                         &mut scratch_2,
                     );
+                    round_glwe_columns_to_k(self, &mut full_ct, 0..1, &mut scratch_2);
```
Pro7ech: I think it would be simpler and more future proof to update the sampling API instead of adding a patch everywhere sampling is called, allow for masking of the output. It's already sampling in the base2k domain, so low-bit masking (or sampling in a scaled smaller base2k) shouldn't be an issue.

Pro7ech (changes requested):

poulpy-ckks/src/test_suite/mul.rs
jdumezy: It is not strictly needed, was used for diagnostic as it was the original problem that lead to the fix.

poulpy-core/src/default/encryption/compressed/gglwe.rs
```diff
@@ -162,6 +162,7 @@ where
                     self.fill_glwe_mask_from_seed_default(base2k, &mut full_ct, 1, rank_out, seed);
                     self.glwe_encrypt_sk_internal(
                         base2k,
+                        full_ct.k().as_usize(),
```
Pro7ech: This is already obtainable by NoiseInfos

poulpy-core/src/default/encryption/glwe.rs
```diff
@@ -15,6 +16,31 @@ use poulpy_hal::{
     source::Source,
 };
 
+fn round_vec_znx_columns_to_k<BE, M>(
```
Pro7ech: Remove

Pro7ech (changes requested):

poulpy-core/src/default/encryption/glwe.rs
```diff
@@ -277,7 +281,9 @@ where
             * (self.bytes_of_vec_znx_dft(1, size) + self.bytes_of_vec_znx_big(1, size) + BE::bytes_of_vec_znx(self.n(), 1, size));
         let lvl_3: usize = self.vec_znx_big_normalize_tmp_bytes();
 
-        lvl_0 + lvl_1 + lvl_2 + lvl_3
+        (lvl_0 + lvl_1 + lvl_2 + lvl_3)
+            .max(self.vec_znx_rsh_tmp_bytes())
```
Pro7ech: This is an artifact from the old rounding

poulpy-core/src/default/encryption/glwe.rs
```diff
@@ -368,6 +374,8 @@ where
         + ScalarZnxFillBinaryHwSourceBackend<BE>
         + ScalarZnxFillBinaryProbSourceBackend<BE>
         + ScalarZnxFillBinaryBlockSourceBackend<BE>
+        + VecZnxRshAssignBackend<BE>
+        + VecZnxLshAssignBackend<BE>
```
Pro7ech: remove

poulpy-core/src/default/encryption/glwe.rs
```diff
@@ -480,6 +489,19 @@ where
                 self.vec_znx_copy_backend(&mut res.data, i, &ci_ref, 0);
             }
         }
+        let padding = res
```
Pro7ech: Why do we need padding in PK encryption?

Pro7ech (approved):

PR #231, opened 2026-08-27
Jp/dev key helper

Pro7ech:
# Reuse GGLWE keys across precision

## Summary

This PR lets GGLWE evaluation keys be reused at coarser decompositions, instead of requiring a separate key for each precision. A key stored with digit size `D` can now be used at any multiple of `D` by selecting the matching rows from the original key.

It also changes high-level APIs so callers provide a key source, not one fixed key. Operations ask that source for the automorphism or tensor key they need at the precision they will actually use. This lets one key set support CKKS provide optimal keys for operations happening at different levels.

## What Changed

- Added selected-row extraction for prepared matrices in `poulpy-hal`, with reference, AVX2, AVX-512, NEON, and Rayon support.
- Added `GetAutomorphismKey` and `GetTensorKey` helpers in `poulpy-core` so callers can choose how keys are stored and resolved.
- Added `GGLWEPrepared::with_dsize` and related layout metadata so a prepared GGLWE key can be viewed at a coarser digit size.
- Updated core operations to take concrete prepared key backend views instead of generic prepared-key type parameters. This is a change was triggered by the key wrapper implementing the same API for different types, requiring full qualified calls (e.g. `to_backend_ref()`)
- Updated CKKS multiplication, squaring, polynomial, approximation, EvalMod, slot-product, and coefficient/slot conversion paths to resolve keys through the new key-source APIs.
- Added `ckks_conjugate_rotate_into`, which performs the fused conjugate-and-rotate operation needed by PaCo.
- Removed the separate conjugation-key plumbing. Rotation keys now provide conjugation through the same automorphism-key source.
- Updated bin-FHE call sites to pass key views through the new signatures.

## Fixes

- Fixed cross-radix tensor relinearization so the DFT operand width comes from the actual input precision instead of the storage precision.
- Made missing CKKS automorphism-key errors report the precision used for the lookup.
- Made `ckks_mul_tmp_bytes` size its tensor temporary from operand precision, instead of the destination's pre-call precision.

## Breaking Changes

- Core evaluation-key operations now take backend key views such as `&GGLWEPreparedBackendRef`, `&GLWEAutomorphismKeyPreparedBackendRef`, `&GGSWPreparedBackendRef`, and `&GGLWEToGGSWKeyPreparedBackendRef`.
- CKKS ciphertext-ciphertext operations now take tensor-key sources and resolve keys at their working precision.
- Coeffs-to-slots and slots-to-coeffs entry points now take one automorphism-key source, including the conjugation element.
- `GLWEAutomorphismKeyHelper`, `CKKSAtkBounds`, separate conjugation-key fields, and the extra generic prepared-key bounds on operation signatures were removed.


poulpy-ckks/src/default/paco/bootstrap.rs
```diff
@@ -224,50 +292,20 @@ where
...
-    validate_gadget_backend_view(
-        "PaCo tensor key",
-        keys.tensor_key(),
-        &tensor_view,
-        plan.n(),
-        context.base2k(),
-        working_size,
-    )?;
-
+    // Every key is checked once, at the branch's working precision. A key that
```
jdumezy, line 295: I think PaCo checks keys only at the initial working precision, while later stages resolve them again after precision decreases. Maybe we should validate and size keys at every precision reached by the schedule?

poulpy-ckks/src/delegates/mul.rs
```diff
@@ -35,6 +51,9 @@ where
         B: CKKSCtBounds,
         T: GGLWEInfos,
     {
+        // Sizing takes the key, not a decomposition: which effective `dsize` the
```
jdumezy, line 54: Scratch sizing uses the stored tensor-key layout, but execution may resolve a coarsened view requiring additional row-extraction space. Could sizing resolve the key at the operation’s actual precision?

poulpy-core/src/default/linear_transformation/baby_steps.rs
```diff
@@ -199,21 +198,25 @@ pub(super) fn glwe_prepare_linear_transformation_baby_steps<BE, M, A, H, K>(
...
-    let has_nonzero_rotation = cache.values.keys().any(|&rot| rot != 0);
-    let (use_hoisted, key_size) = if has_nonzero_rotation {
-        let key_infos = keys.automorphism_key_infos();
-        (
-            a.base2k() == key_infos.base2k(),
-            gglwe_product_output_size::<BE, _, _, _>(a, a, &key_infos),
-        )
-    } else {
-        (false, a.size())
+    // Baby rotations rotate the source, so their keys are the ones the source's
```
jdumezy, line 203: The first baby-step key determines the hoisted path and sizing, although later rotations may resolve keys with different layouts. This decision should probably be derived from all keys used by the transform?

poulpy-core/src/default/linear_transformation/prepared_giants.rs
```diff
@@ -190,28 +188,32 @@ pub(super) fn glwe_eval_giant_steps<BE, M, R, P, H, K>(
...
-            &key_infos,
-            nonzero_giant_rotations,
-            prod_size.saturating_sub(res.size()),
-        );
-        (bases_match, output_size)
-    } else {
-        // No giant rotation: BIG-flow accumulator is always valid (no key required).
-        (true, res.size())
-    };
+    // Giant rotations rotate the post-product destination, so their keys are
```
jdumezy, line 190: The giant-step path and output size are based only on the first key, but subsequent keys are resolved independently.

poulpy-ckks/src/default/paco/bootstrap.rs
Pro7ech: That isn't needed. If a key is valid at precision $K$ then it will also be at precision $k < K$, including for the different dsizes.

poulpy-ckks/src/delegates/mul.rs
Pro7ech: Coarsening is part of the caller-owned key lookup policy. Callers that resolve a coarsened key at a given operation precision should use the same resolved key view when calling *_tmp_bytes. The coarsened prepared view carries the stride, so scratch sizing includes the selected-row extraction buffer. The issue is rather than GGLWELayout should allow to described coarsed keys. I'll add the stride argument to make it complete.

Pro7ech: Now the user can resolve with 

```rust
let key_layout = stored_layout.at_dsize(selected_dsize)?;
let scratch_bytes = module.glwe_tensor_relinearize_tmp_bytes(&res, &tensor, &key_layout);
```

poulpy-core/src/default/linear_transformation/baby_steps.rs
Pro7ech: I agree, fixing that

poulpy-core/src/default/linear_transformation/prepared_giants.rs
Pro7ech: Addressed as the same time as the one for the babies

poulpy-cpu-avx512/src/ntt3x42_ifma/vmp.rs
```diff
@@ -1130,3 +1130,36 @@ fn vmp_apply_dft_to_dft_digits_strided_ifma_inner<E: TaskExecutor>(
...
+    let cols_in: usize = a.cols_in();
+    let (res_rows, res_nrows, res_ncols) = (res.rows(), res.rows() * cols_in, res.cols_out() * res.size());
+    let (a_nrows, a_ncols) = (a.rows() * cols_in, a.cols_out() * a.size());
+    let n_blk_quads: usize = n / 8;
+    // Same strides as the prepare kernel above, per matrix shape.
+    let (res_bq, res_col) = (res_ncols * res_nrows * 16, res_nrows * 16);
+    let (a_bq, a_col) = (a_ncols * a_nrows * 16, a_nrows * 16);
+    let span: usize = cols_in * 16;
+
+    let src: &[u64] = cast_slice(a.raw());
```
jdumezy: The following test fails:

```bash
RUSTFLAGS='-Ctarget-feature=+avx512f,+avx512ifma,+avx512vl' \
      cargo test -p poulpy-cpu-avx512 \
      --features enable-avx512f,enable-ifma,enable-ckks \
      ntt3x42_ifma::tests::ntt3x42_ifma_tests::vmp::test_vmp_extract_selected_rows -- --exact
```

Result: panic because 147,456 bytes are requested from a 98,304-byte packed buffer. The kernel should probably access the underlying packed bytes directly, as the other IFMA kernels do, to avoid the 16 bytes/24 bytes mismatch.

poulpy-core/src/default/linear_transformation/baby_steps.rs
```diff
@@ -224,25 +234,16 @@ pub(super) fn glwe_prepare_linear_transformation_baby_steps<BE, M, A, H, K>(
...
-                        .get_automorphism_key(module.galois_element(rot))
-                        .unwrap_or_else(|| panic!("missing automorphism key for baby-step rotation {rot}"));
-                    (key.p(), key.to_backend_ref())
-                })
-            })
-            .collect();
         let mut tasks: Vec<_> = cache.values.iter_mut().map(|(&rot, prepared)| (rot, prepared)).collect();
         let workers = worker_count::<BE::TaskExecutor>(BABY_ROTATION_WORKERS, tasks.len());
-        let key_infos = keys.automorphism_key_infos();
+        let product_bytes = key_refs
```
jdumezy: I think the public scratch query still underallocates for heterogeneous keys. `glwe_prepare_linear_transformation_baby_steps_tmp_bytes(a, key)` sizes one key using that key’s own output width, but execution first selects the maximum output width across all resolved keys and then evaluates every key’s product scratch at that shared width.

I reproduced this with two valid resolved keys: one reports 32000 bytes and a coarsened key reports 29376 bytes. Allocating max(...) = 32000 panics at the scratch split on line 250.

poulpy-cpu-avx512/src/ntt3x42_ifma/vmp.rs
Pro7ech: Yes, this stems from the change in layout, exactly a 1.5x mismatch. I unfortunately don't always have access to a machine with avx512, thanks for testing it.

jdumezy (approved):

PR #244, opened 2026-08-28
Jdmz/ntt pairwise dsize fusion

jdumezy:


Pro7ech (approved):
Good one 💪 Hope it can also apply to GPU

PR #246, opened 2026-08-29
Canonicalize partial limbs in linear transforms

jdumezy:
Fixes a but where the observed precision in the first few slots was significantly lower than in the rest of the slots after a linear transform.


poulpy-core/src/api/linear_transformations.rs
```diff
@@ -97,6 +97,7 @@ pub trait GLWELinearTransformations<BE: Backend> {
     fn glwe_eval_linear_transformation_into<R, P, H>(
         &self,
         cnv_offset: usize,
+        res_k: usize,
```
Pro7ech: Already available on R. Besides I think there is something wrong if the correction of the issue requires to modify the public API.

Pro7ech (changes requested):
I think the solution is over complicated. I'd suggest simply adding an API on HAL to canoncalize VecZnx, similar to the normalization API. I believe canonical representation deserves at least a flag, so I working on a more permanent solution to be deployed in parallel to the normalization state compile time error. In the mean time, manually calling the canonicalization in the linear transformation using the new HAL API is a good temporary solution.

poulpy-ckks/src/test_suite/linear_transformation.rs
```diff
@@ -57,6 +60,33 @@ fn complex_diagonals<F: TestScalar>(diag_indices: &[usize], m: usize) -> Complex
     ComplexDiagonals::new(re, im)
 }
 
+fn assert_partial_limb_canonical<BE, C>(ct: &C)
```
Pro7ech: Remove this test as the path for canonicalization is not in HAL (which should have its own tests).

poulpy-core/src/default/linear_transformation/baby_steps.rs
```diff
@@ -110,7 +112,14 @@ where
     let hoisted = hoisted_a_dft + scratch_workers::<BE::TaskExecutor>(BABY_ROTATION_WORKERS) * hoisted_worker;
 
     let fallback = baby + module.glwe_automorphism_tmp_bytes(a_infos, a_infos, key_infos).max(prepare);
-    hoisted.max(fallback).max(prepare)
+    let prepare_babies = hoisted.max(fallback).max(prepare);
+    let needs_canonicalization =
```
Pro7ech: Simpler: add a flag in VecZnx: `is_canonical: bool` and a method `.is_canonical() -> bool`

poulpy-core/src/default/linear_transformation/baby_steps.rs
```diff
@@ -110,7 +112,14 @@ where
...
+        !a_infos.k().as_usize().is_multiple_of(a_infos.base2k().as_usize()) || a_infos.size() < a_infos.max_size();
+    if needs_canonicalization {
+        baby + prepare_babies.max(module.vec_znx_canonicalize_tmp_bytes())
```
Pro7ech: Why does canonicalization need tmp bytes?

poulpy-core/src/default/linear_transformation/lazy.rs
```diff
@@ -221,4 +222,22 @@ pub(super) fn glwe_normalize_big_into<BE, M, R>(
             &mut scratch.borrow(),
         );
     }
+    for col in 0..cols {
+        module.vec_znx_canonicalize(res_base2k, res_k, &mut res_ref.data, col, &mut scratch.borrow());
+    }
+}
+
+pub(super) fn glwe_canonicalize<BE, M, R>(module: &M, res: &mut R, scratch: &mut ScratchArena<'_, BE>)
```
Pro7ech: I don't think canonicalization at the core and above should be something public and exposed by the user. I see it as a maintenance operation that should be handled automatically by ops that would "uncanoncalize" canonical inputs.

There aren't many of those (essentially keyswitch and cnv) so the canonicalization should be surgically introduced there as a post processing op. Instead of having to preemptively canonicalize inputs everywhere this could result in a precision degradation.

poulpy-cpu-ref/src/reference/vec_znx/canonicalize.rs
```diff
@@ -0,0 +1,50 @@
+use poulpy_hal::layouts::{Backend, HostDataMut, VecZnxBackendMut, ZnxView, ZnxViewMut};
+
+#[inline]
+fn split_digit(base2k: usize, value: i128) -> (i64, i128) {
```
Pro7ech: Since canonicalization performs a full limb overpass, wouldn't it be better at this point to merge canonicalization with normalization? This would also simplify the model with value certified being normalized being also canonical.

Pro7ech (approved):

PR #254, opened 2026-09-01
Make BinFHE key preparation and execution device-generic

jdumezy:


poulpy-bin-fhe/src/blind_rotation/algorithms/cggi/key_prepared.rs
```diff
@@ -72,3 +75,343 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Pro7ech: What is this test for?

jdumezy: It is a regression test for device-generic key preparation. I agree it is too long for what it does. It can be reduced or removed.

Pro7ech: Yea, its most of the PR size. I wonder why such a test is needed in the first place, what it brings that cross backends in poulpy-hal or poulpy-core doesn't provide.

jdumezy: I'll remove it. I think the difference is that they (cross backends in poulpy-hal or poulpy-core) test transfers between distinct backend types, not truly opaque/device storage.

Pro7ech (approved):

PR #255, opened 2026-09-01
Jp/move normalize flag

### LINKED ISSUE #212: Move `Unormalized` state to poulpy-hal
poulpy-ckks currently has an experimental `Unormalized` state that enforce compile time invalid inputs for ops that MUST take as input normalized values. This works well and should be generalized by moving that to poulpy-hal to automatically apply it to the entire workspace.

Pro7ech:
# Move the `Unnormalized` state to `poulpy-hal`

Fixes #212.

The carry-normalization state, previously CKKS-only, is now a type parameter `VecZnx<D, W, S: NormalizationState = Normalized>` in `poulpy-hal`, propagated through `GLWE` and `CKKSCiphertext`. Feeding un-normalized digits to a DFT-domain primitive is a compile error.

## Rules

- Carry-producing ops require `Unnormalized` destinations, bound-preserving ops map an input state into any state it `FitsIn`, normalizing ops accept any destination, DFT-side inputs require `Normalized`.
- `into_unnormalized` is the free relabel; `normalize(self, ..)` is the only way back. No `assume_normalized` exists anywhere in the safe layer.
- Backend implementors get one explicit bypass for fused kernels: `poulpy_hal::oep::SetNormalizationState` (safe `set_unnormalized()`, unsafe `set_normalized()`). Never used in scheme code.

## Fixes surfaced by the types

Explicit normalizations added in encryption, tensor products, packing, prepared-giants and the bin-FHE CMUX/`FheUint` paths; the standard CGGI blind rotation now normalizes its accumulator per iteration.

Make it faster by explicit override in backends!


poulpy-cpu-arm/src/ntt4x30/rayon.rs
jdumezy, line 1: ARM Rayon backend no longer compiles. There is an omission of the newly required vmp_extract_selected_rows producing E0046 with enable-rayon. It should forward to the base NEON implementation, like AVX Rayon.

poulpy-hal/src/layouts/vec_znx.rs
```diff
@@ -69,30 +72,173 @@ impl VecZnxShape {
...
+impl<D: Data, W: ZnxWord, S: NormalizationState> VecZnx<D, W, S> {
+    /// Relabels this vector as [`Unnormalized`].
+    ///
+    /// Free and always sound: normalized digits are valid unnormalized digits.
+    /// This is the entry point of a fusion loop; [`Self::normalize`] is its exit.
+    ///
+    /// Relabelling a *borrowed view* (rather than the owner) leaves the owner's
+    /// label untouched, so a caller doing that must normalize the view in place
+    /// before the borrow ends.
+    pub fn into_unnormalized(self) -> VecZnx<D, W, Unnormalized> {
```
jdumezy: It seems that the normalization type-state can be bypassed safely. A caller can relabel a borrow, perform carry-producing operations, drop it, then use the unchanged Normalized owner in a DFT operation. Public raw mutattion paths such as data and as_scalar_znx_mut have the same issue. This defeats the advertised compile-time guarantee and can siletly corrupt results.

poulpy-core/src/default/encryption/glwe.rs
jdumezy, line 1: New normalization passes are missing from scratch-size contracts, notably public-key encryption, which advertises only vec_znx_big_normalize_tmp_bytes. No traitx guarantees ordinary normalization needs less scratch than big normalization. The tensor formulas have the same omission in poulpy-core/src/default/operations/glwe.rs. A valid custom backend using the exact advertised allocation can therefore most likely panic.

poulpy-cpu-arm/src/ntt4x30/rayon.rs
Pro7ech: It actually predates this PR, compilation check for NEON wasn't in the CI job. I'll fix it in this PR.

PR #257, opened 2026-09-07
fix #256 & full review pass + optimization on normalization

### LINKED ISSUE #256: Cross base normalization doesn't fully normalize

Pro7ech:
- **new:** CPU-only overwrite and fused extraction hooks with scalar defaults (`I64NormalizeOps` is required when reusing the shared CPU loops), exact integer-oracle tests (595,210 boundary cases across usual and extreme base widths by default; the full 285,149,916-case sweep is ignored), SIMD parity coverage, and cross-base Criterion benchmarks (`--bench normalize`).
- **changed:** Fix [#256](https://github.com/poulpy-fhe/poulpy/issues/256) (cross-base limb ranges), same-base and cross-base truncation rounding and offsets, and wide add/sub normalization; validate extraction slice lengths before writing; reduce memory passes, block serial i64 cross-base calls, and retain vector kernels when same-base truncation drops at most one limb.
- **removed:** Duplicated wide assignment loops, obsolete helpers and redundant output zeroing.
- **others:** Keep HAL backend agnostic: CPU fusion policy, kernels and parity tests live in `poulpy-cpu-ref`; document input bounds, add test-only `dashu-int`, and restore NEON Rayon VMP forwarding.
- **defered:** Native AVX-512/IFMA/NEON validation and further reduction of cross-base normalization overhead.

## Correctness fixes

- **[#256](https://github.com/poulpy-fhe/poulpy/issues/256), cross-base output limbs outside the normalized range:** combining signed sub-digits could leave their sum outside `[-2^(res_base2k-1), 2^(res_base2k-1))`, even when it represented the correct value. For example, an accumulated `-9` at `res_base2k=4` must become digit `7` with carry `-1` into the next more significant limb. The fix centers every completed output limb and propagates its carry before advancing, including partial final limbs and limb 0 before returning; carry beyond limb 0 is discarded modulo one. CPU implementations can perform this centering in the last extraction pass. This covers shared vector/coefficient loops and the wide NTT4x30 copies.

- **Incorrect rounding when dropping precision:** the old loops could round discarded limbs successively and return the wrong nearest value. With `a_base2k=res_base2k=2`, input `[0,1,2]` represents `3/32`; reducing it to one output limb at offset 0 incorrectly returned `[1]` (`1/4`) instead of `[0]`. The exact bit reader now handles cross-base truncation and same-base truncation dropping more than one limb, propagating carries exactly and rounding once at the destination precision, with ties toward positive infinity. Same-base calls dropping at most one limb keep the existing kernels because they already perform only one rounding step. Integer-oracle regressions check the value independently; limb-bound tests separately check the normalized range.

- **Short source or carry slices could cause out-of-bounds SIMD access:** narrow and wide extraction now validate lengths before any writes on reference, AVX2, AVX-512, IFMA and NEON backends, including Rayon forwarding.

## Performance

Against **original `origin/main` (`9361627e`)**, the recorded 4,800-case bounded-input sweep (before the slice-length guards and CPU/HAL separation) takes **49.2% more time** geometrically: **+1.8% same-base2k**, **+64.2% cross-base2k**. Big normalization separately:

| Backend | Same-base2k time change | Cross-base2k time change |
|---|---:|---:|
| FFT64 reference | +1.5% | +98.0% |
| FFT64 AVX2 | +2.2% | +87.6% |
| NTT4x30 reference | +1.3% | +15.4% |
| NTT4x30 AVX2 | +5.7% | +7.1% |

## Added lines by purpose

| Area | Added | Removed | Purpose |
|---|---:|---:|---|
| Shared CPU normalization and scalar kernels | 572 | 334 | Exact rounding, limb scheduling, carry propagation, scalar extraction, CPU hooks and coefficient blocking. |
| SIMD kernels | 496 | 51 | Overwrite and fused extraction, wide arithmetic fixes and scalar fallbacks. |
| Traits and backend integration | 519 | 30 | CPU hook implementations and forwarding across reference, SIMD and Rayon backends; HAL contracts stay backend agnostic. |
| Tests | 1,520 | 130 | Integer oracle, exhaustive cases, regressions, CPU kernel parity and coefficient/range checks. |
| Benchmarks | 198 | 18 | Parameter sweep, bounded inputs and the Criterion benchmark target. |
| Documentation | 42 | 0 | Benchmark instructions, API contracts and CHANGELOG. |
| Build configuration | 5 | 1 | Test oracle dependency, lockfile entries and manifest formatting. |
| Other compatibility fix | 16 | 0 | Restore NEON Rayon VMP selected-row forwarding. |
| **Total** | **3,368** | **564** | **Net increase: 2,804 lines.** |

### DIFF
--- a/poulpy-cpu-avx/src/ntt4x30/vec_znx_big.rs
+++ b/poulpy-cpu-avx/src/ntt4x30/vec_znx_big.rs
@@ -1,6 +1,6 @@
 //! Large-coefficient (i128) ring element vector support for [`NTT4x30Avx`](super::NTT4x30Avx).
 //!
-//! The shared `poulpy-hal` NTT4x30 defaults rely on backend-provided `I128BigOps`
+//! The shared `poulpy-cpu-ref` NTT4x30 defaults rely on backend-provided `I128BigOps`
 //! and `I128NormalizeOps` hooks for vectorized i128 operations.
 
 use super::{
@@ -93,6 +93,24 @@ impl BigWordHadamardProduct for NTT4x30Avx {
 }
 
 impl I128NormalizeOps for NTT4x30Avx {
+    fn znx_extract_digit_mul_i128(base2k: usize, lsh: usize, res: &mut [i64], src: &mut [i128]) {
+        assert!(src.len() >= res.len());
+        unsafe { super::vec_znx_big_avx::nfc_extract_normalize_avx2::<true, false>(base2k, lsh, base2k + lsh, res, src, &mut []) }
+    }
+
+    fn znx_extract_digit_addmul_normalize_i128<const OVERWRITE: bool>(
+        base2k: usize,
+        lsh: usize,
+        res_base2k: usize,
+        res: &mut [i64],
+        src: &mut [i128],
+        carry: &mut [i128],
+    ) {
+        assert!(src.len() >= res.len());
+        assert!(carry.len() >= res.len());
+        unsafe { super::vec_znx_big_avx::nfc_extract_normalize_avx2::<OVERWRITE, true>(base2k, lsh, res_base2k, res, src, carry) }
+    }
+
     #[inline(always)]
     fn nfc_middle_step(base2k: usize, lsh: usize, res: &mut [i64], a: &[i128], carry: &mut [i128]) {
         // SAFETY: NTT4x30Avx::new() verifies AVX2 availability at construction time.
@@ -121,25 +139,5 @@ impl I128NormalizeOps for NTT4x30Avx {
         } else {
             nfc_final_step_assign_scalar(base2k, lsh, res, carry);
         }
-    }
-}
-
-impl poulpy_hal::reference::znx::ZnxExtractDigitAddMulI128 for NTT4x30Avx {
-    fn znx_extract_digit_mul_i128(base2k: usize, lsh: usize, res: &mut [i64], src: &mut [i128]) {
-        assert!(src.len() >= res.len());
-        unsafe { super::vec_znx_big_avx::nfc_extract_normalize_avx2::<true, false>(base2k, lsh, base2k + lsh, res, src, &mut []) }
-    }
-
-    fn znx_extract_digit_addmul_normalize_i128<const OVERWRITE: bool>(
-        base2k: usize,
-        lsh: usize,
-        res_base2k: usize,
-        res: &mut [i64],
-        src: &mut [i128],
-        carry: &mut [i128],
-    ) {
-        assert!(src.len() >= res.len());
-        assert!(carry.len() >= res.len());
-        unsafe { super::vec_znx_big_avx::nfc_extract_normalize_avx2::<OVERWRITE, true>(base2k, lsh, res_base2k, res, src, carry) }
     }
 }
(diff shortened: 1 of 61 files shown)


poulpy-cpu-avx/src/ntt4x30/vec_znx_big.rs
```diff
@@ -123,3 +123,20 @@ impl I128NormalizeOps for NTT4x30Avx {
         }
     }
 }
+
+impl poulpy_hal::reference::znx::ZnxExtractDigitAddMulI128 for NTT4x30Avx {
+    fn znx_extract_digit_mul_i128(base2k: usize, lsh: usize, res: &mut [i64], src: &mut [i128]) {
```
jdumezy: It appears the new safe methods enter unchecked kernels without validating source/carry lengths. Passing a four-element destination and one-element source changed three elements outside the source slice before panicking. With a smaller allocation, this becomes out-of-bounds memory access. AVX512/Neon would need the same protection.

Pro7ech: The ARM / AVX512 paths still need to be validated

poulpy-hal/src/reference/znx/mod.rs
```diff
@@ -92,50 +92,108 @@ pub trait ZnxSwitchRing {
...
     fn znx_extract_digit_addmul(base2k: usize, lsh: usize, res: &mut [i64], src: &mut [i64]);
-}
 
+    /// Extracts a shifted digit into `res`, replacing its previous contents.
+    fn znx_extract_digit_mul(base2k: usize, lsh: usize, res: &mut [i64], src: &mut [i64]);
+
+    /// Extracts and adds a digit, then centers `res + carry` in the output base.
+    /// Requires `1 <= res_base2k <= 62`, both accumulated sums to fit in i64,
+    /// and their centered subtraction to fit in i64.
+    fn znx_extract_digit_addmul_normalize<const OVERWRITE: bool>(
```
jdumezy: The new fused extraction methods and FUSE_NORMALIZE switch put backend optimization choices into HAL and require custom backends to accommodate them. Is that intended?

poulpy-cpu-ref/src/reference/vec_znx/normalize.rs
jdumezy, line 1: The test performs 285149916 normalization calls, with bigint checks for each. It exceeded a 60-second isolated run on my machine.

poulpy-cpu-avx/src/ntt4x30/vec_znx_big.rs
jdumezy: I validated the AVX512 and ARM paths. Still need, like for AVX, to validate source and carry lengths before entering SIMD.

poulpy-cpu-ref/src/reference/vec_znx/normalize.rs
Pro7ech: You know it's a lot when it's more than the number of spy TVs from LG in circulation. No srsly, I agree it's a lot, but the aim was to be 100% exhaustive, because #256 wasn't caugth because of non exhaustive tests. I'll see what I can do to reduce that.

Pro7ech: Fixed

jdumezy (approved):

Issue #234, opened 2026-08-28
HAL/OEP architecture cleanup


Pro7ech:
## Summary

This issue tracks a broader architectural cleanup around the HAL and OEP contract, with three equally important goals:

* **Clean up `poulpy-hal`** and restore it to a minimal, well-defined mathematical interface.
* **Reach 100% OEP parity-test coverage**.
* **Rename and reposition `default` implementations as `reference` implementations**.

This is intentionally a broad issue and will eventually be split into smaller, actionable ones.

## Motivation

`poulpy-hal` is progressively accumulating specialized and fused kernels. While these kernels may make sense for a particular backend, they often make much less sense for others.

As a result, the HAL is starting to lose its identity.

The intended separation should be much stricter:

* `poulpy-hal` defines the **minimal set of mathematical operations** required to express the schemes;
* generic crates such as `poulpy-core` and `poulpy-ckks` define **backend-agnostic scheme logic and reference compositions**;
* backends are responsible for **performance engineering**, including kernel fusion, specialized kernels, alternative layouts, and architecture-specific implementations.

Speedups should therefore not drive the API of `poulpy-hal` or the structure of generic crates.

If a backend benefits from replacing several HAL operations with one fused kernel, that optimization belongs in the backend. It should not require promoting that fused operation into the common HAL simply to make the optimization possible.

The generic layers should describe **what is computed**. The backend should decide **how it is computed efficiently**.

## 1. Clean up `poulpy-hal`

`poulpy-hal` should contain the smallest practical set of well-defined mathematical primitives necessary to fully describe the implementation of the supported schemes.

Specialized kernels that exist primarily for performance reasons should be removed from the HAL wherever they can instead be expressed as compositions of more fundamental operations.

The objective is not to prevent specialization. It is to put specialization at the correct layer.

A backend should be free to implement any OEP using:

* fused kernels;
* specialized kernels;
* different intermediate representations or layouts;
* architecture-specific algorithms;
* any other optimization appropriate for that backend.

None of those choices should unnecessarily leak into the generic HAL interface.

## 2. Move `default` to `reference`

Backend-agnostic crates above `poulpy-hal` already contain compositions under `./default`. These are used whenever an OEP is not overridden by the backend.

Calling them `default` understates their role.

They are effectively the **reference implementations** of the corresponding operations: portable compositions of the minimal HAL that define the expected behavior independently of any backend optimization.

[...]

PR #265, opened 2026-09-10
Jp/hal minimal api

Pro7ech:
## Summary

PR0 of the HAL/OEP cleanup (#234). Lays the foundations the later passes build on: window views on `VecZnx`/`VecZnxBig`, a `PrepareHint` on every prepared type, the shared value-model documentation, and removal of the dead api/OEP surface. Delegates become pure forwarding; every precondition is asserted in the backend kernel. No operation changes semantics. 

## Added

- Window views: `VecZnxShape { n_full, cols, coeff_offset, n, limb_offset, limb_step, size }`, `window_coeffs`, `window_limbs`, `from_shape`, `with_shape` on `VecZnx` and `VecZnxBig`; `is_dense`, `assert_dense`. Coefficient-wise ops (`zero`, `copy`, `add`, `sub`, `negate`, `big_from_small`) accept windows.
- `PrepareHint { Reuse, OneShot }` carried by `SvpPPol`, `VmpPMat`, `CnvPVecL`, `CnvPVecR` and taken by `bytes_of_*`, `*_alloc`, `take_*_scratch`, `from_data`.
- Module docs on `poulpy_hal::api` and `poulpy_hal::layouts`: value model, canonical form, limb rule, columns, three mutation classes, exactness classes, precondition policy, scratch rule.
- `test_suite::window` (window ops, outside-window untouched, ring ops reject windows) on all 16 backend types; `test_word_compat_prepare_hint_sizes`.
- `poulpy_cpu_ref::reference::vmp_select::assert_extractable`, called first by all six `vmp_extract_selected_rows` kernels.

## Changed

- **Breaking:** `Backend::bytes_of_{svp_ppol,vmp_pmat,cnv_pvec_left,cnv_pvec_right}` and the api-level `bytes_of_*`/`alloc`/`take_*_scratch` take a `PrepareHint`; `SvpPPol::shape()` returns `SvpPPolShape`.
- **Breaking:** ring ops and transforms (`rotate`, `automorphism`, `mul_xp_minus_one`, `switch_ring`, `dft`, `idft`, prepares, coefficient-input products) panic on a window view. The check lives in the kernels (cpu-ref reference kernels, avx2, avx512, arm, rayon), never in `poulpy-hal/src/delegates`.
- **Breaking:** `poulpy_hal::reference` moved to `poulpy_cpu_ref::reference::znx`; `poulpy-bin-fhe` keeps its two host kernels in `blind_rotation/host_znx.rs`.
- Flat accessors `raw`/`as_ptr` assert density; `at()` is window-aware.
- One-shot prepared operands (`glwe_mul_plain`, `glwe_mul_const`, linear-transformation unprepared rhs) allocate with `OneShot`; sizing and take use the same hint.

## Removed

- 30 unused api methods (coefficient shift/normalize variants, `add_const`, `add_scalar_into`, `sub_scalar`, `automorphism_rotate`, `split_ring`, `merge_rings`, `transpose`, `hadamard_product_scalar_znx`, `*_from_bytes` wrappers, `vec_znx_big_alloc_n`, `ModuleNew::new_with`) and the eight seeded `[u8; 32]` sampler twins. `Source`-based samplers stay until PR3.
- 19 `Hal*Impl` OEP methods behind them, including `HalModuleImpl::Config`/`new_with` and the dead `vec_znx_big_add_normal`.
- Their `hal_defaults` bodies and `hal_impl_*!` forwards, `reference/vec_znx/{merge_rings,split_ring}.rs`, `vec_znx_transpose`, `vec_znx_fill_normal_ref`, six uncalled `ntt4x30_cnv_*_lazy` kernels.

## Other

- Follow-ups per the spec: PR1 rename pass, PR2 window normalization and `_coeff` removal, PR3 samplers to `poulpy-core`, PR4 derived ops as OEP default bodies, PR5 per-op contracts, PR6 closure.

## Line changes by crate

| Crate | + | − | Purpose |
|---|---:|---:|---|
| `poulpy-hal` layouts | 718 | 139 | `VecZnxShape`, window constructors, `PrepareHint`, `SvpPPolShape`, dense asserts on flat accessors, model docs |
| `poulpy-hal` api | 149 | 400 | 30 dead methods and 8 seeded twins removed; `PrepareHint` params; module doc |
| `poulpy-hal` delegates | 42 | 496 | dead delegates removed; hint params; no asserts remain (pure forwarding) |
| `poulpy-hal` oep | 2 | 195 | 19 dead OEP methods removed incl. `HalModuleImpl::Config`/`new_with` |
| `poulpy-hal` test_suite | 663 | 1152 | `window.rs` (+430) and hint-size test added; tests of removed ops deleted; Source-only sampler tests kept |
| `poulpy-hal` reference | 0 | 1111 | `reference/znx` moved out (14 files, see next row) |
| `poulpy-cpu-ref` | 132 | 952 | receives `reference::znx`; dead default bodies and forwards removed; `assert_dense` in 23 kernels; `assert_extractable` helper; dead lazy kernels removed |
| `poulpy-cpu-avx512` | 179 | 142 | hint params; `assert_dense` in ifma and ntt4x30 kernels and Rayon `dft_apply`; window and hint tests registered |
| `poulpy-cpu-avx` | 92 | 59 | hint params; `assert_dense`/`assert_extractable` in ntt4x30 kernels; window tests registered |
| `poulpy-cpu-arm` | 69 | 44 | same as avx, pattern-edited |
| `poulpy-cpu-rayon` | 24 | 20 | hint params; `assert_dense` in the FFT64 Rayon `dft_apply` macro |
| `poulpy-core` | 76 | 74 | hint at every prepared alloc/take; one-shot sites use `OneShot`; sizing matches take |
| `poulpy-bin-fhe` | 77 | 10 | `host_znx.rs` replaces `poulpy_hal::reference` imports |
| `poulpy-ckks`, `poulpy-bench` | 36 | 27 | hint params |
| `CHANGELOG.md` | 9 | 0 | six `poulpy-hal` bullets under `[Unreleased]` |


poulpy-hal/src/layouts/vec_znx.rs
```diff
@@ -16,51 +16,121 @@ use crate::{
...
+    pub const fn scalar_offset(self, col: usize, limb: usize) -> usize {
+        self.block_index(col, limb) * self.n_full + self.coeff_offset
+    }
+
+    /// Restricts the view to coefficients `offset..offset + len` of every visible limb.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `len == 0` or `offset + len > self.n()`.
+    pub const fn window_coeffs(self, offset: usize, len: usize) -> Self {
```
jdumezy, line 126: Window bound arithmetic wraps in release builds. Here, it evaluates offset + len, and the limb constructor similarly uses unchecked addition and multiplication. For example, VecZnxShape::new(8, 1, 2).window_coeffs(usize::MAX, 2) is accepted in release because the sum wraps. The unchecked scalar_offset calculation can then wrap back into the allocation, returning a safe scalar view onto the wrong coefficients. We could use checked arithmetic for validation and shape composition, ideally also in the inherent offset helpers.

poulpy-hal/src/layouts/vec_znx_dft.rs
```diff
@@ -303,6 +324,29 @@ impl<D: Data, W: DftWord, B: Backend<DftWord = W>> VecZnxDft<D, W, B> {
     }
 }
 
+impl<D: Data, W: DftWord, B: Backend<DftWord = W>> VecZnxDft<D, W, B> {
```
jdumezy, line 327: Public DFT retagging exposes unsupported windows. Despite stating that DFT window support is deferred, here it publicly accepts any VecZnxShape. On packed backends, zero() then computes a smaller representation from the visible dimensions and clears data[..byte_len] in poulpy-hal/src/layouts/vec_znx_dft.rs line ~205, ignoring coefficient and limb offsets. A release probe using an IFMA DFT shape windowed to coefficients 4..8 cleared the first 64 bytes instead of rejecting the view. See if that is a blocker or not.

poulpy-cpu-ref/src/reference/vmp_select.rs
```diff
@@ -5,6 +5,49 @@
...
 
+use crate::layouts::{Backend, VmpPMatBackendMut, VmpPMatBackendRef};
+
+/// Rejects a selection the kernel must not be handed: mismatched prepared
+/// shapes, a truncation that widens, or a last row that is outside `a` or
+/// whose index overflows.
+///
+/// Every kernel entry point calls it first, so the kernel may index without
+/// bounds checks in release.
+pub fn assert_extractable<BE: Backend>(
```
jdumezy, line 17: Prepared-to-prepared operations ignore PrepareHint. assert_extractable validates dimensions but not res.hint() == a.hint() here, after which every current extraction kernel copies representation bytes directly. The SVP copy default does the same in poulpy-cpu-ref/src/hal_defaults/svp_ppol.rs line 50. Once a backend gives Reuse and OneShot different layouts -- as this API explicitly permits -- cross-hint copies will place bytes from one representation under metadata naming the other.

poulpy-hal/src/layouts/vec_znx.rs
Pro7ech: Fixed, VecZnxShape::window_coeffs, window_limbs, block_index and scalar_offset now use checked arithmetic in every build. window_coeffs(usize::MAX, 2) on new(8, 1, 2) panics on the overflowing sum instead of wrapping.

poulpy-hal/src/layouts/vec_znx_dft.rs
Pro7ech: Agreed it is a blocker. VecZnxDft::from_shape and with_shape now assert!(shape.is_dense()), so no windowed VecZnxDft can exist until its kernels do. The DFT representation is backend-owned and packed, so a coefficient or limb offset has no meaning in it; zero and the transforms size their work from the visible dimensions, which is exactly what the IFMA probe hit. All existing callers (the Rayon re-tags) pass a dense shape.

poulpy-cpu-ref/src/reference/vmp_select.rs
Pro7ech: Fixed. assert_extractable and all svp_ppol_copy kernels now assert res.hint() == a.hint() (and equal degree) before any byte moves, and the api docs of both operations state the precondition. The check exposed one real caller: the strided key switch in poulpy-core gathered a key into a scratch matrix hardcoded to Reuse; it now takes the scratch with key.data.hint() and sizes for the larger of the two representations (the layout infos do not carry the hint).

jdumezy (approved):

PR #278, opened 2026-09-13
Jp/hal derived defaults

Pro7ech:
# CKKS ciphertexts canonical at their own k

Stacked on `jp/hal-contracts` (PR5). Part of [#234](https://github.com/poulpy-fhe/poulpy/issues/234).

## Summary

A CKKS ciphertext's data is now canonical at the `k` it reports. Every operation test asserts it, and the `mask` parameter of `cnv_prepare_*` that compensated for the violations is removed.

## Why

`ckks_rotate_into` stamped its destination after the key switch, so the result carried noise below its `k`. The convolution consumer rescales by `2^k`, so that noise returned at full magnitude once the mask was off (PaCo error log2 69, bound -14).

## Added

- `assert_canonical_at_k` in the test suite, run by every decryption helper.
- `ckks_set_k_normalized` and `ckks_set_log_delta_normalized`: relabel, then normalize at the new `k` when it dropped.

## Changed

- Every unary CKKS op stamps the destination before the write, with the final metadata.
- `ckks_eval_mod`, the bootstrap pipeline and the functional bootstrap relabel through the new helpers instead of leaving the low bits below the lowered `k`.
- `glwe_lsh`, `glwe_lsh_assign` and `glwe_rsh` normalize at `res.k()`, not at the allocation width.
- Trace noise model gains the per-halving rounding term; parity suite operands are canonical at `k`.

## Removed

- `mask` from `cnv_prepare_left` / `_right` / `_self` through api, OEP, delegates, kernels, suites and call sites; `msb_mask_bottom_limb`; the `*_masked` conversion primitives.

## Other
- Follow-up: make `set_k` unreachable outside normalization at the type level.

| crate | files | + | - |
|---|---|---|---|
| `poulpy-ckks` | 16 | 212 | 67 |
| `poulpy-cpu-avx512` | 11 | 25 | 196 |
| `poulpy-core` | 7 | 115 | 85 |
| `poulpy-cpu-avx` | 7 | 16 | 143 |
| `poulpy-cpu-ref` | 8 | 19 | 129 |
| `poulpy-cpu-arm` | 5 | 4 | 108 |
| `poulpy-hal` | 7 | 17 | 40 |
| `poulpy-bench` | 1 | 3 | 9 |
| `poulpy-cpu-rayon` | 2 | 0 | 8 |
| `CHANGELOG.md` | 1 | 7 | 0 |
| **Total** | **65** | **418** | **785** |


poulpy-hal/src/oep/derived.rs
```diff
@@ -0,0 +1,847 @@
...
+/// of the four default bodies takes one `res_size`-limb `VecZnx` plus the
+/// normalization carry the shift itself is; the plain `vec_znx_lsh` default
+/// needs only the carry, so this over-reports for it, never under-reports.
+#[doc(hidden)]
+pub fn vec_znx_lsh_tmp_bytes_derived<S, BE>(module: &Module<BE>, res_size: usize) -> usize
+where
+    S: HalVecZnxImpl<BE>,
+    BE: Backend,
+{
+    BE::bytes_of_vec_znx(module.n(), 1, res_size) + <S as HalVecZnxImpl<BE>>::vec_znx_normalize_tmp_bytes(module)
```
jdumezy: The formulas add bytes_of_temp + nested_tmp_bytes, but every arena allocation realigns its start. With FFT64Ref, n=4, one limb, vec_znx_lsh_tmp_bytes advertises 128 bytes while vec_znx_lsh_add needs 160 and panics. The same pattern appears in the VMP, RSH, IDFT-normalize, and convolution composite formulas. The first allocation should be rounded to BE::SCRATCH_ALIGN.

poulpy-hal/src/oep/derived.rs
```diff
@@ -0,0 +1,847 @@
...
+    res_col: usize,
+    a: &VecZnxBackendRef<'_, BE>,
+    a_col: usize,
+    scratch: &mut ScratchArena<'_, BE>,
+) where
+    S: HalVecZnxImpl<BE>,
+    BE: Backend,
+{
+    let res_k: usize = ZnxInfos::size(res) * base2k;
+    <S as HalVecZnxImpl<BE>>::vec_znx_normalize(module, res, base2k, res_k, k as i64, res_col, a, base2k, a_col, scratch);
```
jdumezy: Avoid wrapping usize shifts through i64. k as i64 and -(k as i64) reverse direction for large valid values. With k=usize::MAX, an input coefficient of 2 at radix 4 produces 1 for LSH and 4 for RSH instead of zero. This contradicts the documented beyond-precision behavior and affects the whole shift family, including rsh_assign.

jdumezy (approved):

Issue #266, opened 2026-09-10
Add support for sparse polynomials


Pro7ech:
Sparse polynomials are polynomials in $$Y = X^{N/n}$$ for some divisor of $$N$$. They do not need to be stored in a full `VecZnx` of size $$N$$, only of size $$n$$. Associated operations (such as convolution) can be greatly speedup in those cases, which notably includes sparse CKKS plaintexts.

PR #281, opened 2026-09-13
Jp/hal contracts

Pro7ech:
# PR5: operation contracts on the HAL api

Stacked on `jp/hal-derived-defaults` (PR4). Implements step 5 of the design in `docs/superpowers/specs/2026-09-09-hal-minimal-api-design.md`, for [#234](https://github.com/poulpy-fhe/poulpy/issues/234).

## Summary

Every one of the 113 public traits in `poulpy-hal/src/api` now carries a structured contract beside its prose, 132 blocks in all (`Convolution` is documented per method). A test reads the sources and holds each trait to its block. Writing the contracts turned up one kernel that contradicted the limb rule and seven operations with no pinning test; both are fixed here.

## Contract format

```text
op         vec_znx_add(res, res_col, a, a_col, b, b_col)
class      basis
mutation   out-of-place
domain     res, a, b: VecZnx or windows of one, read at one shared base2k
ensures    res[res_col] = a[a_col] + b[b_col] limb by limb
exact      exact
test       test_vec_znx_add_matches_reference
```

`op`, `class`, `mutation`, `domain`, `ensures`, `exact` and `test` are required; `definition`, `requires`, `fallback`, `override` and `sparse` appear where they say something.

- `class`: `basis` (its own definition, required), `variant` (a basis definition under a different storage or mutation contract, still required), `derived` (an OEP default body a backend may override), `support` (allocation, byte sizes, scratch reports).
- `sparse` records which operand slots will accept a degree other than the module's, per section 4.5, with the substitution they stand for. No behaviour change: the kernels still assert the module degree until PR7 ([#266](https://github.com/poulpy-fhe/poulpy/issues/266)).

## Added

- `poulpy-hal/tests/contracts.rs`: parses `src/api/*.rs`, requires a block per `pub trait`, checks the lines each class needs, and resolves every name on a `test` line against `src/test_suite`. A `test` of `none` is allowed only on `class support`.
- Seven pinning tests in `test_suite`, each against a host oracle as well as across backends, registered on every backend: `test_vec_znx_dft_apply` (the `step` / `offset` limb selection), `test_vec_znx_dft_zero`, `test_vec_znx_big_from_small`, `test_vec_znx_big_inner_sum`, `test_vec_znx_big_col_weighted_sum`, `test_vec_znx_scalar_product`, `test_vmp_zero`.
- A `Contract blocks` and a `Degree embedding` section in the `api` module docs.

## Changed

- **Fix, behaviour:** `poulpy-cpu-ref`'s FFT64 `vec_znx_dft_apply` left a destination limb untouched when its source index `offset + j * step` fell past the input, so the limb kept whatever the buffer held. It now zeroes it, as the NTT4x30 and IFMA kernels already did. Only a non-zero `offset` reaches it; `test_vec_znx_dft_apply` fails without the fix.
- `test_suite::vec_znx_dft::test_vec_znx_copy` is renamed `test_vec_znx_dft_copy`: it tests `vec_znx_dft_copy`, and the old name collided with the coefficient-domain test.
- The api prose no longer names operands the signatures do not have. The `VecZnxBig` add and sub family documented `a`, `b` and `c` where the methods take `res` and `a`; `vec_znx_big_sub_negate_assign` described the wrong direction (it is `res = a - res`). The rotate and automorphism docs name the parameter they take.

| crate | files | + | - | what |
|---|---|---|---|---|
| `poulpy-hal` | 14 | 2397 | 35 | contract blocks, module docs, enforcement test, seven pinning tests |
| `poulpy-cpu-avx512` | 4 | 28 | 0 | test registrations |
| `poulpy-cpu-avx` | 2 | 14 | 0 | test registrations |
| `poulpy-cpu-arm` | 2 | 14 | 0 | test registrations |
| `poulpy-cpu-ref` | 2 | 10 | 1 | the `dft_apply` fix, test registrations |
| `CHANGELOG.md` | 1 | 4 | 0 | Unreleased entries |
| **Total** | **25** | **2467** | **36** | |


poulpy-hal/src/api/vec_znx_dft.rs
```diff
@@ -18,6 +38,16 @@ pub trait VecZnxDftBytesOf {
 ///
 /// The `step` and `offset` parameters select which limbs of the input
 /// are transformed: limbs `offset, offset + step, offset + 2*step, ...`.
+///
+/// ```text
+/// op         vec_znx_dft_apply(step, offset, res, res_col, a, a_col)
+/// class      basis
+/// mutation   out-of-place
+/// domain     res: a VecZnxDft; a: a dense VecZnx of the module degree; step >= 1
```
jdumezy, line 46: The AVX/AVX-512 NTT dft_apply and dft_copy kernels accept step == 0, while reference/IFMA reject it. I reproduced NTT4x30Avx::vec_znx_dft_apply(0, …) completing successfully. This contradicts both the contract and the policy that kernels assert parameter preconditions, so could be worth fixing

poulpy-hal/src/api/mod.rs
```diff
@@ -79,6 +110,17 @@
 //! a * s`, `idft(vmp_apply(dft(a), prep(M))) = a * M`, and the DFT-domain
 //! `add`, `sub`, `automorphism` are the images of the ring operations.
 //!
+//! # Degree embedding
+//!
+//! Every operation is defined on `R_N`. A degree-`n` object, `n` a
```
jdumezy, line 115: nitpick: the new rule says every non-sparse operand has degree N, but vec_znx_big_inner_sum, vec_znx_big_col_weighted_sum, and vec_znx_scalar_product are explicitly coefficient-wise, non-ring operations. LWE encryption routinely uses arbitrary-degree buffers followed by an n = 1 reduction. Following the new global rule would reject those valid calls.

poulpy-hal/src/api/convolution.rs
```diff
@@ -211,11 +415,36 @@ pub trait Convolution<BE: Backend> {
...
     /// Prepares both left and right convolution operands from the same input polynomial,
     /// sharing the FFT/NTT computation. This is an optimization for self-convolution
     /// (squaring) where both operands are the same polynomial.
+    ///
+    /// ```text
+    /// op         cnv_prepare_self(left, right, a, mask, scratch)
+    /// class      derived
+    /// mutation   out-of-place
+    /// definition cnv_prepare_left(left, a, mask) and cnv_prepare_right(right, a, mask)
+    /// domain     left: a CnvPVecL; right: a CnvPVecR; a: a dense VecZnx of the module degree; mask as for cnv_prepare_left
```
jdumezy: cnv_prepare_left/right require their column count to equal a.cols(), and cnv_prepare_self additionally requires equal left/right sizes and columns. None appears in the structured domains. I reproduced the supposedly permitted left.size() = 1, right.size() = 2 call panicking in FFT64. Equal sizes are especially essential because the scratch API accepts only one res_size I think.

poulpy-hal/src/api/vec_znx_big.rs
```diff
@@ -123,8 +243,19 @@ pub trait VecZnxBigSubSmallA<B: Backend> {
     );
 }
 
+/// ```text
+/// op         vec_znx_big_sub_small_assign(res, res_col, a, a_col)
+/// class      variant
+/// mutation   in-place
+/// definition vec_znx_big_sub_small_a(res, res_col, res, res_col, a, a_col) with the operands in the big/small order the signature fixes
```
jdumezy: There is swapped small/big variant definitions. vec_znx_big_sub_small_assign is defined using sub_small_a, but computing big - small requires sub_small_b. The written composition is in fact not type-correct I think. The inverse mistake occurs at line 298 for sub_small_negate_assign.

poulpy-hal/src/api/convolution.rs
```diff
@@ -192,12 +371,37 @@ pub trait Convolution<BE: Backend> {
...
     /// See [Convolution::cnv_apply_dft] for information about the bivariate convolution.
+    ///
+    /// ```text
+    /// op         cnv_pairwise_apply_dft(cnv_offset, res, res_col, a, b, i, j, scratch)
+    /// class      derived
+    /// mutation   out-of-place
+    /// definition cnv_apply_dft(a[i], b[i]) then cnv_apply_dft_add of (a[i], b[j]), (a[j], b[i]) and (a[j], b[j]); i == j degenerates to the single product
+    /// domain     res: a VecZnxDft; a: a CnvPVecL; b: a CnvPVecR; i, j column indices
+    /// requires   scratch >= cnv_pairwise_apply_dft_tmp_bytes(cnv_offset, res.size(), a.size(), b.size())
+    /// ensures    idft(res[res_col]) is the convolution of (a[i] + a[j]) with (b[i] + b[j]), expanded in the DFT domain where the prepared operands are linear
```
jdumezy: The ensures line unconditionally promises (a[i] + a[j]) * (b[i] + b[j]), which becomes 4·a[i]*b[i] when the indices match. The implementation, prose, and definition instead deliberately return one product for i == j. Could be confusing.

poulpy-hal/src/api/vec_znx_dft.rs
```diff
@@ -216,6 +418,19 @@ pub trait VecZnxDftAutomorphism<B: Backend>: VecZnxDftAutomorphismPlan<B> {
...
+    ///
+    /// ```text
+    /// op         vec_znx_dft_automorphism(p, res, res_col, a, a_col)
+    /// class      derived
+    /// mutation   out-of-place
+    /// definition vec_znx_dft_automorphism_with_plan(vec_znx_dft_automorphism_plan(p), res, res_col, a, a_col)
+    /// domain     res, a: VecZnxDft of the module degree; p odd
+    /// ensures    as for vec_znx_dft_automorphism_with_plan, with the plan built and dropped inside the call; prefer the plan form when one p is reused
+    /// fallback   trait default body: build the plan, then apply it
+    /// override   allowed, though building a plan is the cheap half
```
jdumezy: vec_znx_dft_automorphism says a backend may override it, but it is an API trait default and has no HalVecZnxDftImpl OEP method. The generic Module<B> implementation prevents downstream backends from replacing that body.

jdumezy: vec_znx_dft_automorphism says a backend may override it, but it is an API trait default and has no HalVecZnxDftImpl OEP method. The generic Module<B> implementation prevents downstream backends from replacing that body.

jdumezy (approved):

PR #285, opened 2026-09-14
Jp/ckks canonical k

Pro7ech:
# CKKS ciphertexts canonical at their own k

Stacked on `jp/hal-oep-single-backend` Part of [#234](https://github.com/poulpy-fhe/poulpy/issues/234).

## Summary

A CKKS ciphertext's data is now canonical at the `k` it reports. Every operation test asserts it, and the `mask` parameter of `cnv_prepare_*` that compensated for the violations is removed.

## Why

`ckks_rotate_into` stamped its destination after the key switch instead of before and normalization didn't use the correct `k`, so the result carried noise below its `k`.

## Added

- `assert_canonical_at_k` in the test suite, run by every decryption helper.
- `ckks_set_k_normalized` and `ckks_set_log_delta_normalized`: relabel, then normalize at the new `k` when it dropped.

## Changed

- Every unary CKKS op stamps the destination before the write, with the final metadata.
- `ckks_eval_mod`, the bootstrap pipeline and the functional bootstrap relabel through the new helpers instead of leaving the low bits below the lowered `k`.
- `glwe_lsh`, `glwe_lsh_assign` and `glwe_rsh` normalize at `res.k()`, not at the allocation width.
- Trace noise model gains the per-halving rounding term; parity suite operands are canonical at `k`.

## Removed

- `mask` from `cnv_prepare_left` / `_right` / `_self` through api, OEP, delegates, kernels, suites and call sites; `msb_mask_bottom_limb`; the `*_masked` conversion primitives.

## Other

- Follow-up: make `set_k` unreachable outside normalization at the type level.

| crate | files | + | - |
|---|---|---|---|
| `poulpy-ckks` | 16 | 212 | 67 |
| `poulpy-cpu-avx512` | 11 | 25 | 196 |
| `poulpy-core` | 7 | 115 | 85 |
| `poulpy-cpu-avx` | 7 | 16 | 143 |
| `poulpy-cpu-ref` | 8 | 19 | 129 |
| `poulpy-cpu-arm` | 5 | 4 | 108 |
| `poulpy-hal` | 7 | 17 | 40 |
| `poulpy-bench` | 1 | 3 | 9 |
| `poulpy-cpu-rayon` | 2 | 0 | 8 |
| `CHANGELOG.md` | 1 | 7 | 0 |
| **Total** | **65** | **418** | **785** |


poulpy-core/src/default/operations/glwe.rs
jdumezy, line 1: Removing the prepare mask changes a previously enforced behavior into a caller precondition, but GLWEMulPlain still says the ciphertext is processed at its effective k, and GLWETensoring does not state the canonicality requirement. Core’s GLWE type does not enforce that invariant. I reproduced two k = 3 ciphertexts differing only in the unused bit producing different multiplication results (2 versus 4). The affected public operations should most likey either normalize/assert or explicitly require every convolution input to be canonical at its declared k.

poulpy-core/src/default/operations/glwe.rs
```diff
@@ -1816,7 +1798,9 @@ where
         + VecZnxRshTmpBytes
         + VecZnxLshTmpBytes
         + VecZnxLshAssign<BE>
-        + VecZnxLsh<BE>,
+        + VecZnxNormalize<BE>
+        + VecZnxNormalizeAssign<BE>
+        + VecZnxCopy<BE>,
 {
     fn glwe_shift_tmp_bytes_default(&self, res_size: usize) -> usize {
```
jdumezy, line 1813: glwe_shift_tmp_bytes_default still only takes the maximum of the HAL left/right-shift scratch sizes. However, GLWE shifts now independently allocate a temporary and/or invoke vec_znx_normalize[_assign]. Shipped backends happen to advertise enough scratch, but the OEP contract permits a backend overriding all shift kernels to advertise less. Such a conforming backend can then panic when given exactly the scratch size returned by the core API.

poulpy-core/src/default/operations/glwe.rs
```diff
@@ -1834,9 +1818,35 @@ where
...
+        // copy back, so the result is canonical at the `k` it reports.
+        // `glwe_shift_tmp_bytes` already reserves that temporary.
+        let n: usize = self.n();
+        for i in 0..cols {
+            let (mut tmp, mut scratch_iter) = scratch.borrow().take_vec_znx_scratch(n, 1, res.size());
+            self.vec_znx_normalize(
+                &mut tmp,
+                base2k,
+                res_k,
+                -(k as i64),
```
jdumezy: The new partial-width glwe_rsh and out-of-place glwe_lsh paths use k as i64. For k > i64::MAX, this wraps and can turn an enormous shift into a small shift instead of producing zero. I reproduced usize::MAX returning nonzero results for both operations. This is the same conversion issue recently fixed in HAL’s derived shifts.

poulpy-core/src/default/operations/glwe.rs
Pro7ech: A normalized ciphertext is also now expected to be canonical at its `k`. Two `k=3` ciphertexts that differ in a unused bit are not normalized, so they sit outside of the contract. the GLWE layout now documents this behavior and ops calling cnv say they expect normalized operands and do not check. I agree that eventually we want to have a compile time enforcement of this contract, but this is for future work.

poulpy-core/src/default/operations/glwe.rs
Pro7ech: Fixed and tested.

poulpy-core/src/default/operations/glwe.rs
Pro7ech: Same bug as above, and same fix, and tested.

poulpy-core/src/default/operations/glwe.rs
```diff
@@ -1816,10 +1805,19 @@ where
...
+        + VecZnxNormalizeAssign<BE>
+        + VecZnxNormalizeTmpBytes
+        + VecZnxCopy<BE>,
 {
     fn glwe_shift_tmp_bytes_default(&self, res_size: usize) -> usize {
-        self.vec_znx_rsh_tmp_bytes(res_size).max(self.vec_znx_lsh_tmp_bytes(res_size))
+        // The partial-width `glwe_rsh` path normalizes into a one-column
+        // temporary, and `glwe_lsh_assign` normalizes in place after the
+        // kernel; both need the normalization scratch on top of the kernels'.
+        let normalize: usize = BE::bytes_of_vec_znx(self.n(), 1, res_size) + self.vec_znx_normalize_tmp_bytes();
```
jdumezy: The new scratch formula adds BE::bytes_of_vec_znx(...) directly to the normalization requirement, but ScratchArena aligns the next allocation to BE::SCRATCH_ALIGN. When the temporary size is not already aligned, that padding consumes bytes the formula does not reserve. A conforming backend with smaller HAL shift scratch requirements can therefore panic when given exactly glwe_shift_tmp_bytes() (reproduced experimentally). We could probably wrap the temporary size in BE::scratch_aligned(...) before adding vec_znx_normalize_tmp_bytes().

Pro7ech: fixed

jdumezy (approved):

PR #289, opened 2026-09-15
Jp/hal hot path

Pro7ech:
# Hot paths: one by-const kernel, Rayon lanes in CI

On `jp/hal-closure` (PR6). Part of [#234](https://github.com/poulpy-fhe/poulpy/issues/234). One by-const kernel replaces four, hot loops shed their per-coefficient checks, CI gains the Rayon lanes.

## Changed

- The four `cnv_by_const_apply` / `_add` kernels become one `poulpy-cpu-ref` kernel over `BigWord = i128`; NTT backends forward to it, FFT64 keeps its own.
- `NTT4x30Avx512` `cnv_by_const_apply` at 4096x8: 1,162 us to 220 us (pre-cleanup `main`: 704 us); `ckks_mul_pt_const_into`: 48.0 us to 14.6 us.
- Five other hot loops lose per-coefficient checks: `div_round_*`, the `normalize_exact` fallbacks, `cnv_apply_dft_sum`, the `mat_vec` bbc products.
- The by-const contract lets `b` have any degree with `b_coeff < b.n()`, as `poulpy-ckks` already calls it; the kernel checks that in debug builds.
- `enable-rayon` joins the AVX, AVX-512 and NEON CI feature sets; the NEON test step now uses `$NEON_FEATURES`, so its Rayon suites run.

## Other

- `ZnxView::at_ptr` is unchanged: a safe accessor called once per limb, and no coefficient loop reaches it anymore.
- `poulpy-cpu-arm` with `enable-rayon` was only `cargo check`-able here (no cross-linker); its Rayon suites run for the first time on the NEON lane.

| crate | files | + | - |
|---|---|---|---|
| `.github` | 1 | 5 | 1 |
| `CHANGELOG.md` | 1 | 8 | 4 |
| `poulpy-cpu-avx` | 3 | 62 | 102 |
| `poulpy-cpu-avx512` | 5 | 42 | 230 |
| `poulpy-cpu-ref` | 4 | 128 | 91 |
| `poulpy-hal` | 2 | 5 | 4 |


poulpy-cpu-ref/src/reference/ntt4x30/vec_znx_big.rs
```diff
@@ -1408,25 +1415,27 @@ pub fn ntt4x30_vec_znx_big_normalize_assign<O, R, A, BE>(
     let mut output = res.to_backend_mut();
     let output_size = output.size();
     assert!(carry.len() >= 3 * output.n());
-    for i in 0..output.n() {
+    let input_limbs: Vec<&[i128]> = (0..input.size()).map(|j| input.at(a_col, j)).collect();
+    let out_limbs: Vec<*mut i64> = (0..output_size).map(|j| output.at_mut(res_col, j).as_mut_ptr()).collect();
```
jdumezy: Each output.at_mut() reborrows the entire backing buffer, invalidating previously cached pointers. With multiple output limbs, the subsequent writes use invalid pointers. Miri reports undefined behavior on a valid two-limb input, the same input passes on the base revision. Taking one base pointer and deriving limb addresses from the shape fixes the reproduction.

poulpy-cpu-ref/src/reference/ntt4x30/convolution.rs
```diff
@@ -901,50 +909,66 @@ fn ntt4x30_cnv_by_const_apply_impl<BE, E: TaskExecutor, const ADD: bool>(
...
-    let a_size = a.size();
-    let b_size = b.size();
+    let (res_size, a_size, b_size) = (res.size(), a.size(), b.size());
+    let n = res.n();
+    let res_cols = res.cols();
+    debug_assert!(
+        res_col < res_cols,
+        "ntt4x30_cnv_by_const_apply_impl: res_col {res_col} >= cols {res_cols}"
+    );
+    debug_assert!(a.n() == n, "ntt4x30_cnv_by_const_apply_impl: a.n() {} != res.n() {n}", a.n());
```
jdumezy: The new loop stops at the shorter of the input and output slices. If the input degree is smaller, it silently leaves part of the output unchanged. Previously, this invalid call panicked. The new debug_assert! only catches it in debug builds. I think we can use assert_eq!(a.n(), res.n()) before the loop so release builds reject it too.

jdumezy (approved):

PR #290, opened 2026-09-15
Jp/hal pointer provenance

Pro7ech:
# One pointer per slice in the FFT butterflies and the reim conversions

## Changed

- Twelve FFT/IFFT butterflies (AVX2, AVX-512, NEON) called `as_mut_ptr()` up to four times per slice; each call invalidates the pointers before it.
- The three in-place `reim_to_znx_i64_assign_bnd63_*` conversions took two typed pointers from two `res.as_mut_ptr()` calls on the same slice.
- Every kernel now takes one pointer per slice and derives its offsets from it; results are bit-identical, suites pass unchanged.

| crate | files | + | - |
|---|---|---|---|
| `CHANGELOG.md` | 1 | 1 | 0 |
| `poulpy-cpu-arm` | 2 | 33 | 25 |
| `poulpy-cpu-avx512` | 3 | 35 | 26 |
| `poulpy-cpu-avx` | 3 | 35 | 26 |


jdumezy:
LGTM, one issue that I missed before that should be fixed at some point in poulpy-cpu-avx/src/fft64/reim/fft_avx2_fma.rs, line109. _mm_load_pd requires 16-byte alignment, but omg: [f64; 2] only guarantees 8. Miri reproduces undefined behavior. The same issue exists in poulpy-cpu-avx/src/fft64/reim/ifft_avx2_fma.rs, line 105. _mm_loadu_pd should be used.

jdumezy (approved):

PR #292, opened 2026-09-15
Jp/hal sparse operands

### LINKED ISSUE #266: Add support for sparse polynomials

Pro7ech:
# Sparse operands: degree-n add/sub and a sparse convolution right operand

This PR has a little bit more than anticipated and includes some cleaning of the HAL contracts.

## Changed

- `VecZnx` / `VecZnxBig` add and sub take a degree-n operand in every sparse-capable slot: one strided kernel set behind the reference kernels.
- A right operand prepared under a degree-n module feeds every apply form of a degree-N module through the slot law `slot_N(i) = slot_n(i >> log2(N/n))`.
- The prepares are not sparsity-aware and unchanged; the result and the left operand take the module degree, asserted once at every apply entry.
- Sparse right operand on every family: one row load and one lane permute per row (AVX FFT64, IFMA), per-slot broadcast loads (AVX NTT4x30), scalar (ref, NEON).
- Scratch: FFT64 apply grows by `8 * b_size` f64, NTT4x30 reference by the staged `b` rows, IFMA by `32 * b_size` u64; IFMA sum sized as max(fused, unfused).
- HAL contracts: the `exact` key is gone, results are exact under their operand bounds; no contract or rustdoc names a backend, crate, ISA, spec section or PR.
- Convolution contracts state one truncation bound, `a.size() + b.size() - 1 - cnv_offset` limbs, and one degree rule per operand; no rationale prose.

## Other

- Rulings: add/sub stay scratch-free (no default body); degree floor 8 checked at the apply entry; rank-1 tensor kernels take the module degree.
- Parity vs `switch_ring` then dense, on every backend: `test_vec_znx_sparse_add_sub`, `test_vec_znx_big_sparse_add_sub`, `test_convolution_sparse`.

| plaintext degree | FFT64Avx512 | NTT4x30Avx512 | NTT3x42Ifma |
|---|---|---|---|
| 2^16 (dense) | 247 us | 2038 us | 1247 us |
| 2^14 | 186 us, 1.33x | 1635 us, 1.25x | 984 us, 1.27x |
| 2^12 | 163 us, 1.52x | 1544 us, 1.32x | 920 us, 1.36x |
| 2^10 | 160 us, 1.54x | 1523 us, 1.34x | 907 us, 1.38x |

| crate | files | + | - |
|---|---|---|---|
| `CHANGELOG.md` | 1 | 3 | 1 |
| `poulpy-cpu-arm` | 2 | 12 | 0 |
| `poulpy-cpu-avx512` | 8 | 292 | 73 |
| `poulpy-cpu-avx` | 5 | 148 | 29 |
| `poulpy-cpu-rayon` | 1 | 41 | 15 |
| `poulpy-cpu-ref` | 13 | 562 | 115 |
| `poulpy-hal` | 30 | 943 | 292 |


jdumezy:
The updated contract (poulpy-hal/src/api/vec_znx_big.rs, line 107) enables sparse inputs, but the fallback (poulpy-hal/src/oep/derived.rs, line 463) calls vec_znx_big_from_small, which requires equal degrees. Reproduced with degree-8 b and degree-64 res: the CPU override succeeds but the fallback panics. Both subtraction fallbacks have the same problem.

jdumezy (approved):

PR #298, opened 2026-09-16
Jp/hal definitions

Pro7ech:
Every HAL contract carries a `definition` line, the precise mathematical statement of the operation, in a notation the `api` module docs define once. 

## Changed
- Every `basis`, `variant` and `derived` block gets a `definition` line, a formula or a composition of named HAL operations; `support` blocks carry none, the contract test enforces both.
- `# Notation` defines `a[c,j,i]` indexing, `old(a)`, `canon(v,b,k,L)`, `wrap`, `idft`, prepared operands read as their sources, `reseed(source)`, `omega(m,r)`.
- The `exact` key and the exact/approximate DFT class are gone; every result is exact under its domain bounds.
- Corrections found while deriving the formulas: convolution scale `2^((cnv_offset + 1) * w)`, vmp column equalities and zeroed tail, normalize `res_k` bound, right-shift `+ 64` bound, sampler on the reseeded stream, no fallback carve in `_tmp_bytes` ensures.
- `NegacyclicFFT::fft` and `ifft` are basis operations with an ideal complex definition, pinned by `test_negacyclic_fft` on every FFT table.
- HAL contract lines carry no backend name, citation or rationale prose.
- Rustdoc prose is one sentence per trait and one per method; argument lists, diagrams and restatements of the block are gone.
- The `vmp_apply_dft_to_dft` column preconditions are asserted by every kernel (NTT4x30 reference, AVX2, AVX-512, IFMA, NEON; the Rayon overrides call them), pinned by `test_vmp_apply_dft_to_dft_shape_rejected` on every backend.
- The contract test holds every trait method to exactly one block whose `op` line names it, so a deleted or missing contract fails.

| crate | files | + | - |
|---|---|---|---|
| CHANGELOG | 1 | 2 | 0 |
| poulpy-cpu-arm | 1 | 14 | 0 |
| poulpy-cpu-avx | 1 | 14 | 0 |
| poulpy-cpu-avx512 | 1 | 14 | 0 |
| poulpy-cpu-ref | 3 | 17 | 60 |
| poulpy-hal | 14 | 848 | 759 |


jdumezy:
The new VMP shape preconditions are not enforced by optimized NTT backends. The contract now requires exact column equality in poulpy-hal/src/api/vmp_pmat.rs line 173, but only the reference NTT kernel gained assertions. Optimized AVX code still flattens mismatched inputs without checking in poulpy-cpu-avx/src/ntt4x30/vmp.rs line 394, silently computing a differently shaped product instead of panicking. AVX-512, IFMA, Neon, and Rayon overrides follow the same pattern, including the accumulating variant.

jdumezy:
The contract test does not seem to enforce “one definition per computing operation.” poulpy-hal/tests/contracts.rs line 207 only verifies that each trait has at least one contract, then validates whichever blocks happen to exist. For multi-method traits such as NegacyclicFFT and Convolution, deleting a method’s contract (or adding an undocumented computing method)still passes. The checker should maybe associate each op with an actual trait method and require coverage for every computing method. Or this can be differed for later.

jdumezy (approved):

PR #307, opened 2026-09-18
Jp/dft diagonal sparsity

Pro7ech:
## Summary

Every diagonal of a `Diagonals` map is stored as one period of its values; the S2C and C2S factors are generated that way, layer by layer, and encoded on that many slots. On the full-packing presets the four S2C factors go from degree N each to 16, 256, 4096 and N (C2S mirrored), and the matrices shrink the same way.

## Why

Each radix-2 layer is a block-diagonal butterfly: its three diagonals repeat over the butterfly width, a merge keeps the longer period, so every diagonal of a factor repeats over the widest butterfly it merges and its plaintext is `M(X^(2^s))`. The compact-plaintext encoder stored every diagonal at `2 * slots`.

## Added

- `Diagonals::period(index)` (core); `ComplexDiagonals::diagonal_period`, `period`, `scale`.
- `DFTPlan::diagonal_log_sparsity` / `factor_log_sparsity`: the same rule replayed from the schedule, value-free.
- Tests: stored periods, replay and measured periods agree per diagonal; compiled factor degrees per factor; compact maps evaluate, compose and encode like their dense replication.

## Changed

- **Breaking (core):** `Diagonals::set` takes one period (a length dividing `slots`); `get` returns it; readers index modulo it; `rotate_slots_into` reads its source as periodic.
- The DFT generator builds layers one butterfly wide and merges compact vectors; `build_transform` hands the encoder the map's period, so the existing compact allocation does the rest. Dense maps are unchanged.

## Other

- Sparse-repack seed and half-zeroed last `Encode` factor keep only the packing sparsity; bit-reversed layers count at their full width.

| crate | files | + | - |
|---|---|---|---|
| poulpy-core | 1 | 69 | 12 |
| poulpy-ckks | 5 | 481 | 132 |
| CHANGELOG | 1 | 2 | 0 |


poulpy-ckks/src/layouts/complex_diagonals.rs
```diff
@@ -132,20 +176,23 @@ impl<T: DiagonalArithmetic> ComplexDiagonals<T> {
...
+                let period_b = rhs.diagonal_period(ib).expect("indexed diagonal present");
                 let idx = (ia + ib).rem_euclid(t);
-                let mut re = out.re.get(idx).cloned().unwrap_or_else(|| zeros.clone());
-                let mut im = out.im.get(idx).cloned().unwrap_or_else(|| zeros.clone());
-                for j in 0..t {
-                    let ar = read(&self.re, ia, j, ta);
-                    let ai = read(&self.im, ia, j, ta);
-                    let br = read(&rhs.re, ib, j + ia, tb);
-                    let bi = read(&rhs.im, ib, j + ia, tb);
+                let len = out.diagonal_period(idx).unwrap_or(1).max(period_a).max(period_b) as i64;
```
jdumezy: Maybe an edge case: diagonals::set accepts any period dividing the slot count. For six slots, composing periods [1,2] and [10,20,30] produces [10,40,30,10,40,30] instead of [10,40,30,20,20,60]. The combined period is LCM(2,3)=6, not 3. The same assumption appears in diagonal_period() and period().

jdumezy (approved):

PR #308, opened 2026-09-18
Jp/hal aligned buf

Pro7ech:
## Summary

This PR adds Miri to `poulpy-hal` CI and it flagged a few things, which are addressed here.

`alloc_aligned` allocated at 64-byte alignment and returned a `Vec` that freed at the element alignment, a layout mismatch the allocator contract forbids and Miri reported on the first drop of any layout. `poulpy_hal::AlignedVec<T>` (`AlignedBuf = AlignedVec<u8>`) now owns that storage and frees it with the layout that made it; every owned layout, every host backend's `OwnedBuf` and the twiddle tables are built on it. A `miri` CI job runs the `poulpy-hal` unit tests.
## Why

A `Vec` cannot carry the layout it was allocated with, so the fix is an owner type with `Drop` and one mechanical sweep of the `Vec<u8>` owned-container spelling.

## Added

- `AlignedVec<T: Copy>`: zero-initialized, stored `Layout`, `Deref<[T]>`, `Clone`, `Eq`, `Hash`, `Default`, `truncate`, `From<&[T]>` / `From<Vec<T>>` (copy into 64-byte-padded storage, zero tail). Unit tests run under Miri.
- CI job `miri`: `cargo miri test -p poulpy-hal --lib`, 42 tests, under 10 s locally; serialization tests run at degree 16 and the huge-page advice is skipped under Miri.

## Changed

- **Breaking:** `alloc_aligned` / `alloc_aligned_custom` return `AlignedVec<T>`; `VecZnxOwned`, `MatZnxOwned`, `ScalarZnxOwned`, every host backend `OwnedBuf` and every owned scheme layout (`GLWE<AlignedBuf, W>` and siblings) are spelled on `AlignedBuf`.
- **Breaking:** layout `from_bytes` takes `impl Into<AlignedBuf>` (a `Vec<u8>` still compiles and is copied); `Backend::from_bytes` copies instead of adopting; `ToOwnedDeep` / `to_host_owned` copy into aligned storage.
- The NTT and reim twiddle tables (`powomega`, `omg`) hold `AlignedVec`.

## Removed

- `alloc_aligned_custom_u8` and its "CRITICAL-2" note; the `assert_alignment` calls in `from_bytes` (the type guarantees it); two no-op `resize` lines in the IFMA tables.

## Other

- Verified: fmt, clippy and tests on the portable, AVX and AVX-512 lanes, aarch64 lib clippy, contracts, rustdoc, Miri.
[...]

PR #311, opened 2026-09-18
Jp/hal ckks followups

Pro7ech:
# hal, ckks: follow-up bundle after the aligned owner type and the compact plaintexts

## Summary

The leftovers recorded by the #308 and #305 reviews: `Backend::from_bytes` gets a default body, the callerless alignment asserts go, `AlignedVec::clone` allocates for its length, a layout's `from_bytes` accepts an unpadded byte count, the transfer tests poison before they copy; CKKS prepares a right operand at its own degree and `ckks_extract_pt` reports a degree mismatch; the core `oep` docs link the reference modules.

## Why

Ten identical `from_bytes` bodies, two dead `pub fn`, a clone that kept a truncated buffer's whole allocation, an equality on the unpadded `bytes_of` that only held because every in-tree size is a 64-byte multiple, tail-zero legs that passed on a fresh allocation alone, two right slots taken at the module degree that a compact plaintext does not need, a kernel panic where a composition error belongs, and eighteen dead intra-doc links.

## Added

- `AlignedVec::capacity`; `poulpy_hal::layouts::padded_bytes`.
- Tests: the clone of a truncated buffer, a layout built from an unpadded byte count, `decrypt_extract_degree_mismatch_error` on every CKKS backend.

## Changed

- `Backend::from_bytes` default body `Self::from_host_bytes(&bytes)`; the nine backends, the host backend, the delegating macro and the core test doubles drop theirs.
- `AlignedVec::clone` allocates `len` padded to its alignment; the `Layout` expect names the size overflow.
- The seven layout `from_bytes` compare against `padded_bytes(bytes_of)`.
- `test_transfer_padded_lengths` fills the destination with `0xff` before `copy_from_host` and `copy_host_to_view`.
- `ckks_prepare_right` allocates at `a.n()`; the SHIP masking accumulation takes each right slot at `pi.n()` (the scratch bound stays at the module degree).
- `ckks_extract_pt` returns `PlaintextDegreeMismatch` when the plaintext degree differs from the ciphertext's.
- The core `oep` docs link `crate::reference::*` modules instead of eighteen names that never existed (rustdoc errors on `poulpy-core` 49 to 31, the rest predate this stack).
- The docs state the #234 model everywhere the layering is described (root, core, ckks and backend READMEs, the getting-started guide, CONTRIBUTING, the crate docs, the `reference` and `oep` module docs, the HAL `oep` docs): `reference` is the implementation, the only validated circuit; an override is a faster route to the same result, pinned by the parity suites, never a different behaviour.

## Removed

- **Breaking:** `poulpy_hal::assert_alignment` and `assert_alignement`.

## Other

- Deferred: the NEON permute gather (needs the arm host to measure), butterfly sparsity under sparse repack (design), the import drift of the #308 sweep (cosmetic).

| crate | files | + | - |
|---|---|---|---|
| poulpy-hal | 12 | 68 | 46 |
| poulpy-core | 13 | 54 | 49 |
| poulpy-ckks | 10 | 92 | 16 |
| poulpy-cpu-ref | 3 | 3 | 9 |
| poulpy-cpu-avx | 3 | 3 | 9 |
| poulpy-cpu-avx512 | 4 | 3 | 12 |
| poulpy-cpu-arm | 3 | 3 | 9 |
| docs | 3 | 8 | 8 |
| CHANGELOG | 1 | 2 | 0 |


poulpy-hal/src/aligned.rs
```diff
@@ -141,9 +146,13 @@ impl<T: Copy + Zeroable> Default for AlignedVec<T> {
     }
 }
 
+/// The clone allocates for its own length, padded to the alignment, whatever
+/// the original's allocation holds.
 impl<T: Copy + Zeroable> Clone for AlignedVec<T> {
     fn clone(&self) -> Self {
-        let mut out: Self = Self::zeroed(self.layout.size() / size_of::<T>(), self.layout.align());
+        let align = self.layout.align();
+        let size = (self.len * size_of::<T>()).next_multiple_of(align);
```
jdumezy: Cloning a valid truncated buffer can now panic, e.g.:
```rust
let mut v = AlignedVec::<[u64; 3]>::zeroed(8, 64);
v.truncate(1);
let copy = v.clone(); // panics
```
The clone rounds 24 bytes to 64, then divides by 24, requesting two elements (48 bytes) which violates the allocation’s alignment requirement. We should round to a common multiple of the alignment and element size.

jdumezy (approved):

